PR #1, opened 2021-10-21
Subspace testnet resources

1devNdogs:
Updated for subspace testnet.

- Contains just 2 instances.
   - relayer-app-backend (To run Serge relayer backend in typescript)
   - test-rpc-1 (want to replicate the bootnode in here, to document and confirm the installation Justin and Nazar done in the current working test-rpc)

- Also updated readme.

Still, as a draft, need to add instructions for each droplet config. We can discuss this idea and priority later in a sync meet, so will be not started yet.

As an extra comment about terraform features:

- In the future we can add all the docker, configs, nginx, certificates, etc directly to terraform and have these instances fully automated, should take some so for now a .md with instructions must be enough.

- Also we can configure droplets to limit close root access, change default ssh port and allow only auth login with ssh keys and remove user password authentication.


.gitignore
```diff
@@ -1,4 +1,6 @@
 .terraform.lock.hcl
 .terraform.tfstate.lock.info
 .terraform
-out-plan.*
\ No newline at end of file
+current-plan.tfplan
+terraform.tfstate
+terraform.tfstate.backup
```
nazar-pc, line 6: We should really use non-local state storage if we want this to be usable by anyone else.

nazar-pc (approved):

.gitignore
1devNdogs: Ahhh you are rigth, let me see if i get you.
You mean for internal use? so anyone in the team who clone this repo will be able to run the same resources states?

nazar-pc: Yeah. At previous work place we had AWS S3 bucket that had state files, such that everyone on the team could work with it. I guess here we need to include them into the repo instead.

1devNdogs: Ok, i will fix that. Just had to revert one commit.

1devNdogs:
Close PR as this infraestructure will not be used for now.
Will create a separate PR treating this as a not used template and some usefull related docs.

PR #4, opened 2021-11-11
Update docs for latest dev and test stages

1devNdogs:
# Summary.

Only docs updates for #3


nazar-pc (approved):
Skimmed though it, looks fine. But I noticed that Nginx config doesn't enable HTTP2 and exclusively uses outdated TLS versions, only 1.2 and 1.3 (if supported) should be enabled, 1 and 1.1 should be considered broken.

Issue #15, opened 2021-11-22
Update aries-test-a and aries-test-b with same definitions as aries-dev. 


1devNdogs:
# Description

Update **aries-test-a** and **aries-test-b** with same definitions as **aries-dev.** 

## Updates required:

- [x] - **aries-test-a:**  increase volume to 1100gb. Move to nyc3
- [x] - **aries-test-b:**  increase volume to 1100gb and c-8 droplet size (dedicated) Move to nyc3

### We must coordinate how to do this. 

1. Update **aries-test-a** - Currently _**not public.**_
   1.1. Increase volume size. 
   1.2. Reset network and relayer, using the latest images.
   1.3. Wait for full block archive import.

2. Switch **test-rpc** CNAME record to **aries-test-a** leaving **aries-test-b** available to be updated.

3. Update **aries-test-b** - Current _**public**_ env.
_This requires a full droplet reset, cant update from **s-4vcpu-8gb** to **c-8** by disk size conflicts_
   3.1 Increase volume size, Increase droplet size.
   3.2 Update **aries-test-rpc-b** DNS record to new droplet address.
   3.3 Launch network and relayer, using the latest images.

PR #16, opened 2021-11-22
Update dev test env - current state

1devNdogs:
# Description.

These definitions represent the current state of **dev** and **test** resources.

# Changes.
- **aries-dev** updated to **c-8** (dedicated) and volume size increased to **1100gb**.
- **aries-test** removed to work with **a** and **b**
- **aries-test-a** created using **c-8** (dedicated) and 400gb volumen
- **aries-test-b** created using **s-4vcpu-8gb** (shared) and 400gb volumen
- Add team keys to get access after deployment.

## Notes
We have non-consistent resources among dev and test environments since they have been modified as required.
I have created an issue #15  with the required changes so that all environments have the same resources.


resources/aries-test-a/provider.tf
```diff
@@ -0,0 +1,12 @@
...
+}
+data "digitalocean_ssh_key" "nazar-key" {
+  name = "Nazar SSH Key"
+}
+data "digitalocean_ssh_key" "leo-key" {
+  name = "Leo SSH Key"
+}
+data "digitalocean_ssh_key" "serge-key" {
+  name = "Serge SSH Key"
+}
```
nazar-pc, line 12: Would be nice to de-duplicate these in the future somehow

nazar-pc (approved):

ImmaZoni (approved):

Issue #21, opened 2021-12-13
Create new environment for farmer-net


1devNdogs:
# Description.

Create a new env for farmer network. Include a-b for blue green deployment.

_Only bootnode, farmer and rpc nodes._

- [x] Cloudflare dns: farm-rpc.subspace.network (also A and B)
- [x] Terraform definition 
- [x] Deploy and config

PR #24, opened 2021-12-13
Add resources for farmers network

1devNdogs:
# Description

- Resources definitions for : #21 

Will also create a new DNS record called. `farm-rpc.subspace.network`


resources/aries-test-a/droplets.tf
```diff
@@ -9,3 +9,15 @@ resource "digitalocean_droplet" "aries-test-a-nodes-farmer-relayer" {
     data.digitalocean_ssh_key.leo-key.id,
   ]
 }
+
+resource "digitalocean_droplet" "aries-farmers-network-a-nodes" {
+  image  = "ubuntu-20-04-x64"
+  name   = "aries-farmers-network-a-nodes"
+  region = "nyc3"
+  size   = "c-8"
```
nazar-pc: Let's use shared quad-core/8G with regular SSD instead, it really doesn't need to be very powerful for this purpose. And let's run it with stock SSD (no volume) for now, should be enough if we don't have relayer running, we'll add it later if necessary.

nazar-pc (approved):

PR #27, opened 2021-12-22
Update docs, docker images, compose files, certbot, nginx.

1devNdogs:
# Summary

- Update certbot section, add public record on certificate generation. 
- Update `docker hub` references to `ghcr.io`, docs and docker-compose files.
- Add `DATE` reference and docs to replace snapshop-`DATE`.
- Update parameters on node-docker-compose.yml, `public-rpc-node` and `farmer`.

### DIFF
--- a/_docs/5_SubspaceNetwork/README.md
+++ b/_docs/5_SubspaceNetwork/README.md
@@ -109,7 +109,7 @@ server {
   root /var/www/html;
   index index.html;
 
-  location /farmer-rpc {
+  location /farmer {
     proxy_buffering off;
     proxy_pass http://127.0.0.1:9955/;
     proxy_set_header X-Real-IP $remote_addr;
@@ -120,6 +120,30 @@ server {
     proxy_set_header Connection "upgrade";
   }
 
+  location /ws {
+    proxy_buffering off;
+    proxy_pass http://localhost:9944/;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+  }
+
+  location /http {
+    proxy_buffering off;
+    proxy_pass http://localhost:9933/;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+  }
+  
   location / {
     try_files $uri $uri/ =404;
 
--- a/_docs/5_SubspaceNetwork/README.md
+++ b/_docs/5_SubspaceNetwork/README.md
@@ -62,6 +62,17 @@ ln -s /snap/bin/certbot /usr/bin/certbot
 
 Following [certbot instructions](https://certbot.eff.org/lets-encrypt/ubuntufocal-nginx). We need a **subdomain (RPC_PUBLIC_NAME)** mapped to the Droplet **IP_ADDRESS** and a running **nginx** with an **open port 80**.
 
+Environments use a `blue-green` strategy. Creating a new node or environment for public use requires to issue certificates for the <CNAME_record> and <new_subdomain>. Certbot will ask for a domain list to validate. 
+
+For example: (`<new_subdomain>, <CNAME_record>`)
+* Public Testnet: `aries-test-rpc-a.subspace.network, test-rpc.subspace.network` - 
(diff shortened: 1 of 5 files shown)


_docs/5_SubspaceNetwork/node-docker-compose.yml
```diff
@@ -41,12 +41,15 @@ services:
       "--telemetry-url", "wss://telemetry.polkadot.io/submit/ 1",
       "--bootnodes", "/dns/bootstrap-node/tcp/30333/p2p/GENERATED_BOOTSTRAP_NODE_ID_HERE",
       "--rpc-cors", "all",
-      "--rpc-external",
+      "--rpc-methods", "Unsafe",
       "--ws-external",
+      "--validator",
```
nazar-pc: Why this change? I don't think this correct.

isSerge (dismissed):

_docs/5_SubspaceNetwork/README.md
```diff
@@ -98,6 +109,17 @@ server {
...
+  location /farmer-rpc {
+    proxy_buffering off;
+    proxy_pass http://127.0.0.1:9955/;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_http_version 1.1;
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
+  }
```
nazar-pc: Can we have `/farmer` or `/farmer-ws` and add `/ws` and `/http` for node RPC endpoints, so it is consistent.

nazar-pc (changes requested):

_docs/5_SubspaceNetwork/node-docker-compose.yml
1devNdogs: I got them from subspace [docs](https://github.com/subspace/subspace/blob/b3de7b72a427ca51da4cfff37eda4f71c1778fdc/README.md?plain=1#L48), 

This was required on `test-b` restart using `dec-20` snapshots

nazar-pc: But that is for farmer, which this is for public RPC node, they have different requirements and thus different parameters. I really recommend you to run `subspace-node --help` and read what they are so you don't have to guess.

1devNdogs: :confused:  Those instructions are for subspace-node,  I added this because i get this error with latest images deployment:
```
public-rpc-node_1  | Thread 'main' panicked at 'Authority node is required by overseer', crates/subspace-node/src/service.rs:367
```
So i checked the docs and realized that `--validator` was added to run the _node_, after updating this on `public_rpc_node`, i got following error: 
```
public-rpc-node_1  | Error: SubstrateCli(Input("--rpc-external and --ws-external options shouldn't be used if the node is running as a validator. Use `--unsafe-rpc-external` or `--rpc-methods=unsafe` if you understand the risks. See the options description for more information."))
```
In that moment i checked the --help, and they where available, so i changed the `unsafe-rpc-external` to `rpc-methods=unsafe` and it runs. 

Is this a bug then? are those values not expected on **public_rpc_node**??

nazar-pc: Well, if you have to specify something **unsafe** just to make it run and you don't understand why, don't do that, ask someone, there is a good reason why those APIs are unsafe and don't allow node to start, it can be very damaging to deploy random things to production. This is also the reason why I was asking to first create a PR and only after it is reviewed and approved deploying changes to production. If I understand correctly, we already run `--rpc-methods Unsafe` in production, which should have never happened!

@liuchengxu I think node's code needs to be updated to not only check `role.is_authority()` instead of keystore container (which we don't really use anyway). WDYT?

liuchengxu: @nazar-pc That makes sense, the fact that the local keystore exists does not mean the overseer should be enabled, and so `role.is_authority()` is. A new flag `--executor` or `--enable-overseer` is the right way IMO, thoughts?

nazar-pc: I think `--executor` is an eventual goal, I'll update to just use authority check in the meantime.

@1devNdogs le't switch these docs to correct ones and re-deploy with different settings once we have the next snapshot ready.

nazar-pc: https://github.com/subspace/subspace/pull/221

_docs/5_SubspaceNetwork/README.md
1devNdogs: Sure, i updated this, to use /farmer instead and added /http enpoint.

_docs/5_SubspaceNetwork/node-docker-compose.yml
1devNdogs: I took some time to read about this, thanks for the explanations now I understand better the effects of unsafe rpc flag. Sent an update with correct flags.

nazar-pc (approved):

PR #29, opened 2022-01-17
Move dev to nyc

1devNdogs:
Updates to move Dev env to NYC3.

Note: this will not remove the old SFO project as got this issue under investigation.


nazar-pc (dismissed):
Can you explain how this will not remove the old instance though?

1devNdogs:
The old dev env project name is now "aries-dev-old". 
I will be `re-creating` `terraform` state for this project, which will get a new urn and ids not related to the old dev env.
After bootstrapping everything, We should use the old environment to test the pull image issue and remove it as this concludes.
 
Before running this, will coordinate with Serge to avoid interrupting what is running in the environment.

nazar-pc (approved):

PR #45, opened 2022-04-14
add Sora and Turing

isSerge:


_docs/6_BlockArchives/docker-compose.yml
```diff
@@ -437,7 +437,7 @@ services:
     restart: unless-stopped
     environment:
       DEBUG: 1
-      SOURCE_CHAIN_RPC: wss://nodle-parachain.api.onfinality.io/public-ws
+      SOURCE_CHAIN_RPC: wss://nodle-parachain.api.onfinality.io/public-ws?apikey={ONFINALITY_API_KEY}
```
nazar-pc: Are you sure it is public endpoint with token? Previously it was `/ws` endpoint instead.
Also it is a bit non-obvious that you'd need to set an environment variable before running `docker-compose` might be easy to forget.

isSerge: no, it's `public` or `public-ws`, you can check above.

if one forgets to set env variable, then it will just use the public endpoint - nothing will break

nazar-pc: I mean I don't disagree that it may work, but I'm fairly confident that API does nothing. How did you verify it actually does anything and why did you not use `/ws` endpoint as shown in the OnFinality dashboard?

isSerge: I always used `public-ws` and I know that without apiKey it disconnects more often. Why do you think API does not do anything?

nazar-pc: Because this is public API and is not supposed to be used with API token. Ask folks on Telegram to clarify, but even if it works, it works by accident.

isSerge: they are not replying... I can just put `/ws` everywhere

nazar-pc: Yes, let's do that, please

nazar-pc (approved):

PR #47, opened 2022-04-28
feat(relayer): adds additonal relayer droplet

### LINKED ISSUE #46: Relay Net machine
We need a new machine for relayer.
We'll switch from testnet naming to relaynet, so domain will be aries-relay-rpc-a, maybe different domains for bootstrap node URL and WS endpoint as well.

cc @isSerge

Genysys:
This PR adds an additional droplet for the relayer.

This PR closes #46

### DIFF
--- a/resources/aries-test-b/droplets.tf
+++ b/resources/aries-test-b/droplets.tf
@@ -12,9 +12,9 @@ resource "digitalocean_droplet" "aries-test-b-nodes-farmer-relayer" {
   ]
 }
 
-resource "digitalocean_droplet" "aries-test-b-nodes-farmer-relayer-2" {
+resource "digitalocean_droplet" "aries-relaynet-a" {
   image  = "ubuntu-20-04-x64"
-  name   = "aries-test-b-nodes-farmer-relayer-2"
+  name   = "aries-relaynet-a"
   region = "nyc3"
   size   = "c-8"
   ssh_keys = [
--- a/resources/aries-test-b/droplets.tf
+++ b/resources/aries-test-b/droplets.tf
@@ -7,8 +7,6 @@ resource "digitalocean_droplet" "aries-test-b-nodes-farmer-relayer" {
     data.digitalocean_ssh_key.nazar-key.id,
     data.digitalocean_ssh_key.serge-key.id,
     data.digitalocean_ssh_key.leo-key.id,
-    # TODO: Add floating IP so that you can add this without changing the IP address
-    # data.digitalocean_ssh_key.sam-key.id,
   ]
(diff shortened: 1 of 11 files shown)


nazar-pc (dismissed):

resources/aries-test-b/droplets.tf
```diff
@@ -7,6 +7,21 @@ resource "digitalocean_droplet" "aries-test-b-nodes-farmer-relayer" {
     data.digitalocean_ssh_key.nazar-key.id,
     data.digitalocean_ssh_key.serge-key.id,
     data.digitalocean_ssh_key.leo-key.id,
+    # TODO: Add floating IP so that you can add this without changing the IP address
+    # data.digitalocean_ssh_key.sam-key.id,
+  ]
+}
+
+resource "digitalocean_droplet" "aries-relaynet-a" {
```
nazar-pc, line 13: Why do we have `relaynet-a` in `test-b` file?

resources/aries-test-b/outputs.tf
```diff
@@ -3,6 +3,11 @@ output "ip_address_test_droplet" {
   description = "ip_address_test_droplet: "
 }
 
+output "ip_address_test_droplet-2" {
```
nazar-pc: This probably shouldn't be a test either.

resources/farmnet-a/droplets.tf
```diff
@@ -0,0 +1,12 @@
+resource "digitalocean_droplet" "aries-farmnet-a" {
```
nazar-pc, line 1: Looks like this is a new instance we need for farmnet and not relaynet-related, right?

resources/farmnet-a/volumes.tf
```diff
@@ -0,0 +1,12 @@
+resource "digitalocean_volume" "aries-farmnet-a" {
+  region                  = "nyc3"
+  name                    = "farmnet-a-volume"
+  size                    = 1800
+  initial_filesystem_type = "ext4"
+  description             = "Extra volume for relaynet a."
```
nazar-pc: This is not a relaynet though, is it?

resources/aries-test-b/droplets.tf
Genysys: This was an error when it was deployed , and to change this now will bring down the current network , as the server would be redeployed

resources/farmnet-a/droplets.tf
Genysys: yes

nazar-pc (approved):

PR #49, opened 2022-05-09
Feat/relayer refactor

Genysys:
This PR: 

* Refactors the relayer terraform files.
* updates relayer docs.

### DIFF
--- a/_docs/5_SubspaceNetwork/README.md
+++ b/_docs/5_SubspaceNetwork/README.md
@@ -270,7 +270,7 @@ To enable logging to datadog , run the following command:
 
 ```bash
 docker run -d --name datadog-agent \
-    -e DD_API_KEY=fc259acf7ea56abf2596efe89592dae7 \
+    -e DD_API_KEY=<INSERT DATADOG KEY> \
     -e DD_LOGS_ENABLED=true \
     -e DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL=true \
     -e DD_CONTAINER_EXCLUDE_LOGS="name:datadog-agent" \
--- a/_docs/5_SubspaceNetwork/README.md
+++ b/_docs/5_SubspaceNetwork/README.md
@@ -48,7 +48,7 @@ To expose the public RPC node over a secure WebSocket connection.
 apt-get install -y nginx
 ```
 
-### Certbot setup
+### Certbot setup 
 
 To generate an SSL certificate for the public RPC node.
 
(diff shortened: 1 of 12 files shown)


resources/aries-test-b/config.json
```diff
@@ -0,0 +1,275 @@
+{
+    "targetChainUrl": "wss://aries-relaynet-a.subspace.network/",
+    "primaryChain": {
+        "downloadedArchivePath": "/archives/kusama",
+        "wsUrls": [
+            "wss://kusama-rpc.polkadot.io",
+            "wss://kusama.api.onfinality.io/public-ws?apikey=38f60995-1ca8-474f-bbd1-bd80844433ad",
```
nazar-pc, line 7: Putting production secrets like API keys in the repo, especially public, is always a bad idea. I had to revoke these and @isSerge will have to redeploy all relayers now with new keys. These should be removed and replaced with obvious placeholders instead.

resources/aries-test-b/droplets.tf
```diff
@@ -1,18 +1,18 @@
...
-  size   = "c-8"
-  ssh_keys = [
-    data.digitalocean_ssh_key.nazar-key.id,
-    data.digitalocean_ssh_key.serge-key.id,
-    data.digitalocean_ssh_key.leo-key.id,
-    # TODO: Add floating IP so that you can add this without changing the IP address
-    # data.digitalocean_ssh_key.sam-key.id,
-  ]
-}
+# resource "digitalocean_droplet" "aries-test-b-nodes-farmer-relayer" {
```
nazar-pc, line 1: Why not removing it entirely?

resources/aries-test-b/droplets.tf
```diff
@@ -1,18 +1,18 @@
...
+#     data.digitalocean_ssh_key.nazar-key.id,
+#     data.digitalocean_ssh_key.serge-key.id,
+#     data.digitalocean_ssh_key.leo-key.id,
+#     # TODO: Add floating IP so that you can add this without changing the IP address
+#     # data.digitalocean_ssh_key.sam-key.id,
+#   ]
+# }
 
-resource "digitalocean_droplet" "aries-relaynet-a" {
+resource "digitalocean_droplet" "aries-relaynet-test" {
```
nazar-pc, line 15: Why is `aries-relaynet-a` replaced with `aries-relaynet-test`?

resources/aries-test-b/nginx.conf
```diff
@@ -0,0 +1,86 @@
...
+        # server_name_in_redirect off;
+
+        include /etc/nginx/mime.types;
+        default_type application/octet-stream;
+
+        ##
+        # SSL Settings
+        ##
+
+        ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3; # Dropping SSLv3, ref: POODLE
```
nazar-pc, line 35: This should only have `1.3`, maybe `1.2` if we want to be extremely concervative. No 1 and 1.1 for sure.

resources/aries-test-b/testnet/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+    restart: unless-stopped
+    command: [
+      "--chain", "testnet",
+      "--base-path", "/var/subspace",
+      "--wasm-execution", "compiled",
+      "--execution", "wasm",
+      "--pruning", "archive",
+      "--pool-kbytes", "51200",
+      "--telemetry-url", "wss://telemetry.polkadot.io/submit/ 1",
+      "--node-key", "f593e5994721e5c413d40cff0c5a230fdd86f7855197c5bf55230f710be6dbf7",
```
nazar-pc, line 19: This is a TERRIBLE idea, we use this secret key on farmnet and can't change it quickly. Why is it in public repo???

_docs/5_SubspaceNetwork/README.md
```diff
@@ -263,3 +263,21 @@ docker-compose up -d
 
 Typical commands like `docker-compose restart`, `docker-compose logs --tail=100 -f` can be used to manage this setup,
 see [Docker Compose docs](https://docs.docker.com/compose/reference/) for details.
+
+### Datadog Integration
```
nazar-pc, line 267: It is already described in `_docs/4_Docker/README.md` file why do we need it here too?

nazar-pc (changes requested):
I don't fully understand what this does.
The changes that created relay-a (that we need in #46) were reverted in the middle of this branch.
In DigitalOcean I see we have aries-test-test project with aries-relaynet-a in it, which doesn't make sense together.

UPD: I see test-b is commented out here, but I still see it in DigitalOcean for some reason with a DIFFERENT aries-relaynet-a in it and other nodes. Weren't terraform changes applied to the actual infra state?

PR #55, opened 2022-06-15
X-net and Gemini-1b changes

vedhavyas:
- New X-net network

- Gemini-1b changes
	  - Reduce in and out peers to 500
	  - Add extra nodes in singapore
	  - Use latest release

- Migrate from local backend to Terraform cloud backend
- Set max socket connections


resources/gemini-1b/backend.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  backend "remote" {
+    hostname      = "app.terraform.io"
+    organization  = "subspace-sre"
```
ghost, line 4: We can create another organisation for production usage. subspace-sre was created mostly for testing.

resources/gemini-1b/extra_nodes.tf
```diff
@@ -0,0 +1,128 @@
+resource "digitalocean_droplet" "gemini-1b-extra" {
+  count = var.extra-droplets
+  image  = "ubuntu-20-04-x64"
```
ghost: why ubuntu 20.04 and not 22.04?

resources/gemini-1b/extra_nodes.tf
```diff
@@ -0,0 +1,128 @@
...
+    ]
+  }
+
+}
+
+
+# deployment version
+# increment this to restart node with any changes to env and compose files
+locals {
+  deployment_version_extra = 7
```
ghost: How is this used ?

resources/gemini-1b/variables.tf
```diff
@@ -19,5 +19,11 @@ variable "droplets-per-region" {
 variable "node-snapshot-tag" {
   description = "Node snapshot tag"
   type = string
-  default = "gemini-1b-2022-june-05"
+  default = "gemini-1b-2022-jun-10"
```
ghost: Should there be a newer version ? gemini-1b-2022-jun-18 

https://github.com/subspace/subspace/pkgs/container/node/26000223?tag=gemini-1b-2022-jun-18

services/gemini-1/docker-compose-extra.yml
```diff
@@ -0,0 +1,34 @@
...
+      - "30333:30333"
+    command: [
+      "--chain", "gemini-1",
+      "--base-path", "/var/subspace",
+      "--execution", "wasm",
+      "--pruning", "archive",
+      "--pool-kbytes", "51200",
+      "--listen-addr", "/ip4/0.0.0.0/tcp/30333",
+      "--rpc-cors", "all",
+      "--rpc-external",
```
ghost: How port 9933 is used ?

services/x-net/setup_executor_compose.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+cat > /subspace/docker-compose.yml << EOF
+version: "3.7"
+
+services:
+  node:
+    image: vedhavyas/subspace-node:latest
```
ghost, line 8: https://hub.docker.com/layers/subspace-node/vedhavyas/subspace-node/latest/images/sha256-e3f01474fb07c38dc8d378b68afbbe3c05ad20a2684ffa172c7589d4549fff2e?context=explore how's it different from GitHub version ?

services/x-net/setup_rpc_compose.sh
```diff
@@ -0,0 +1,61 @@
...
+cat > /subspace/docker-compose.yml << EOF
+version: "3.7"
+
+volumes:
+  caddy_data: {}
+  archival_node_data: {}
+
+services:
+  # caddy reverse proxy with automatic tls management using let encrypt
+  caddy:
```
ghost, line 12: Maybe it makes sense to add log rotation here 
```

logging:
      driver: "json-file"
      options:
        max-size: "50m"
        max-file: "10"
        
        ```

File: resources/gemini-1b/extra_nodes.tf (hunk 2 above)
Comment (vedhavyas) [reply resources/gemini-1b/extra_nodes.tf]: LTE and most used so far. No opinion on 22 though

File: resources/gemini-1b/backend.tf (hunk 1 above)
Comment (vedhavyas) [reply resources/gemini-1b/backend.tf:4]: is there a difference between this org in comparision to new one? If not,  I think we can delete the testing repos and can create a new one for testing. wdyt ?

File: resources/gemini-1b/extra_nodes.tf (hunk 3 above)
Comment (vedhavyas) [reply resources/gemini-1b/extra_nodes.tf]: This is tag to trigger proviosioner for redeploying docker images without redeploying everything

File: resources/gemini-1b/variables.tf (hunk 4 above)
Comment (vedhavyas) [reply resources/gemini-1b/variables.tf]: good catch. Will need to redeloy those

File: services/x-net/setup_executor_compose.sh (hunk 6 above)
Comment (vedhavyas) [reply services/x-net/setup_executor_compose.sh:8]: this is a testnet that do not have docker images. So I created one for x-net. Also we use ghcr.io instead of docker hub

File: resources/gemini-1b/backend.tf (hunk 1 above)
Comment (ghost) [reply resources/gemini-1b/backend.tf:4]: There's no difference. Sure lets create another one for testing. Also I planning to use them for stages. That can be run straight out of UI by click of the button

File: resources/gemini-1b/extra_nodes.tf (hunk 2 above)
Comment (nazar-pc) [reply resources/gemini-1b/extra_nodes.tf]: 22.04 is also LTE, let's use 22.04 going forward by default, nothing wrong with using it for existing instances.

File: services/gemini-1/docker-compose-extra.yml (hunk 5 above)
Comment (nazar-pc) [reply services/gemini-1/docker-compose-extra.yml]: These nodes don't have RPC at all (or don't need to have at least)

File: services/x-net/setup_executor_compose.sh (hunk 6 above)
Comment (nazar-pc) [reply services/x-net/setup_executor_compose.sh:8]: Doesn't `jun-18` image contain all the changes necessary?

File: services/x-net/setup_rpc_compose.sh (hunk 7 above)
Comment (nazar-pc) [reply services/x-net/setup_rpc_compose.sh:12]: There are some default limits in Docker by default, 250M IIRC, but I don't think it keeps more than 1 file. Log rotation shouldn't be necessary though if we have DataDog for this.

File: resources/gemini-1b/extra_nodes.tf (hunk 3 above)
Comment (ghost) [reply resources/gemini-1b/extra_nodes.tf]: Can we move this logic to Ansible later on ?

File: services/gemini-1/docker-compose-extra.yml (hunk 5 above)
Comment (ghost) [reply services/gemini-1/docker-compose-extra.yml]: so rpc-external flag can be removed, correct ?

File: services/gemini-1/docker-compose-extra.yml [hunk 8]
```diff
@@ -0,0 +1,34 @@
...
+      "--ws-external",
+      "--in-peers", "500",
+      "--out-peers", "250",
+      "--in-peers-light", "500",
+      "--ws-max-connections", "10000",
```
ghost: Is there a patch of a base linux kernel? AFAIK it supports no more than 1000 ws connections per service in default settings

nazar-pc: We had issues with 1000 and no issues once we raised it to 10k so my impression is that is should work, but maybe I'm wrong about it :man_shrugging:

ghost:
```root@gemini-1b-node-extra-9-sfo3:~# sysctl -a | grep soma
net.core.somaxconn = 4096

4k ws allowed

Comment (ghost) [reply services/gemini-1/docker-compose-extra.yml]: root@gemini-1b-node-extra-9-sfo3:~#

Comment (nazar-pc) [reply services/gemini-1/docker-compose-extra.yml]: Then yes, we need to tune this. I'm sometimes wondering why those limits are so low on Linux...

File: services/x-net/setup_executor_compose.sh (hunk 6 above)
Comment (vedhavyas) [reply services/x-net/setup_executor_compose.sh:8]: I think @liuchengxu has more changes coming in. We need to update the images once those docker images are up anyway which may not be part of jun-18. I'll let @liuchengxu weigh in on this

File: services/gemini-1/docker-compose-extra.yml (hunk 8 above)
Comment (vedhavyas) [reply services/gemini-1/docker-compose-extra.yml]: I dont think we need to a kernel patch. I think updating systconf to required connections should be fine in this case

File: resources/gemini-1b/extra_nodes.tf (hunk 3 above)
Comment (vedhavyas) [reply resources/gemini-1b/extra_nodes.tf]: Going forward sure. No opinion this :-D

File: services/x-net/setup_executor_compose.sh (hunk 6 above)
Comment (liuchengxu) [reply services/x-net/setup_executor_compose.sh:8]: Yeah, `jun-18` was enough before, but not now due to this critical executor bugfix https://github.com/subspace/subspace/pull/627 added very recently. I have also built a docker image on ghcr.io in my fork and updated the executor node in the cloud. Additionally, I also have to make another image with [a substrate patch](https://github.com/subspace/subspace/commit/754eca4bfed70931443027fb420f80090283f66b) to help debug https://github.com/paritytech/substrate/issues/11732. Currently, I just ensure the images are fine and replace them directly on the server, happy to have some instructions on the upgrade/patching in a more organized way and follow them later.

File: resources/common.tf [hunk 9]
```diff
@@ -16,7 +16,9 @@ terraform {
 variable "do_token" {}
 
 # SSH agen identity to use to connect to remote host
-variable "ssh_identity" {}
+variable "ssh_identity" {
+  default = "root"
```
nazar-pc: Why is this needed?

resources/gemini-1b/backend.tf
```diff
@@ -4,7 +4,7 @@ terraform {
     organization  = "subspace-sre"
 
     workspaces {
-      name = "gemini-1b"
+      name = "gemini-1b-latest"
```
nazar-pc: What is the purpose of this renaming?

resources/gemini-1b/droplets.tf
```diff
@@ -4,5 +4,10 @@ resource "digitalocean_droplet" "gemini-1b" {
   name   = "gemini-1b-node-${count.index}-${var.droplet-regions[count.index % length(var.droplet-regions)]}"
   region = var.droplet-regions[count.index % length(var.droplet-regions)]
   size   = var.droplet-size
-  ssh_keys = local.ssh_keys
+  ssh_keys = [
+    data.digitalocean_ssh_key.alexei2-key.id,
+    data.digitalocean_ssh_key.nazar-key.id,
+    data.digitalocean_ssh_key.serge-key.id,
+    data.digitalocean_ssh_key.ved-key.id
+  ]
```
nazar-pc: Why? Isn't it equivalent?

resources/gemini-1b/extra_nodes.tf
```diff
@@ -1,128 +1,135 @@
...
-  provisioner "remote-exec" {
-    inline = [
-      "docker compose -f /subspace/docker-compose.yml down --remove-orphans",
-      "echo NODE_SNAPSHOT_TAG=${var.node-snapshot-tag} > /subspace/.env",
-      "docker compose -f /subspace/docker-compose.yml up -d",
-    ]
-  }
-}
-
+#resource "digitalocean_droplet" "gemini-1b-extra" {
```
nazar-pc: Why do we have these files if they are fully commented-out?

resources/gemini-1b/utils/scripts/generate_node_keys.sh
```diff
@@ -0,0 +1 @@
+../../../../scripts/generate_node_keys.sh
```
nazar-pc: Why not just move files here? I don't think they're fully reusable anyway, so copying makes sense.

resources/gemini-1b/projects.tf
```diff
@@ -1,13 +1,14 @@
-resource "digitalocean_project" "gemini-1b" {
-  name        = "gemini-1b"
+resource "digitalocean_project" "tf_refactoring" {
+  name        = "tf_refactoring"
```
nazar-pc: I don't see such project in DO, but I do see empty `gemini-1b` project. If this doesn't correspond to DO state what else doesn't?

resources/gemini-1b/provisioner.tf
```diff
@@ -28,8 +28,9 @@ resource "null_resource" "setup_nodes" {
     host = digitalocean_droplet.gemini-1b[count.index].ipv4_address
     user = "root"
     type = "ssh"
-    agent = true
-    agent_identity = var.ssh_identity
+    agent = false
+#    agent_identity = var.ssh_identity
+    private_key = var.alexey2_do_private_key
```
nazar-pc: Why this change here and below?

nazar-pc (changes requested):

resources/common.tf
ghost: yes, will fix it

resources/gemini-1b/backend.tf
ghost: its needed for terraform cloud

resources/gemini-1b/droplets.tf
ghost: it will be fixed

resources/gemini-1b/extra_nodes.tf
ghost: I can delete it

resources/gemini-1b/utils/scripts/generate_node_keys.sh
ghost: its used for x-net

resources/gemini-1b/projects.tf
ghost: tf_refactoring was used to test tf cloud and refactor current gemini-1b setup. I can switch it back to gemini

resources/gemini-1b/provisioner.tf
ghost: its used for tf cloud

resources/gemini-1b/backend.tf
nazar-pc: Needed how/why?

resources/gemini-1b/extra_nodes.tf
nazar-pc: Then please do

resources/gemini-1b/utils/scripts/generate_node_keys.sh
nazar-pc: Okay, but the question still remains unanswered...

resources/gemini-1b/projects.tf
nazar-pc: I mean if we follow infrastructure-as-code, then code should be identical to infrastructure and vice versa. It is not "I can" kind of situation, it feels like this repo contains some random terraform code that isn't representative of our actual infrastructure. If that is the case we're in a really bad shape because neither `main` nor this PR match the actual infrastructure.

This should be avoided as much as possible, please make sure this PR contains code that **actually** corresponds to the infrastructure we have or else I don't see the point of using terraform and have this repo at all.

resources/gemini-1b/provisioner.tf
nazar-pc: I'd like a bit longer explanation than that. Why does it have to include your private key in variables?

resources/gemini-1b/droplets.tf
```diff
@@ -4,10 +4,5 @@ resource "digitalocean_droplet" "gemini-1b" {
...
   size   = var.droplet-size
-  ssh_keys = [
-    data.digitalocean_ssh_key.alexei2-key.id,
-    data.digitalocean_ssh_key.nazar-key.id,
-    data.digitalocean_ssh_key.serge-key.id,
-    data.digitalocean_ssh_key.ved-key.id
-  ]
-}
+  ssh_keys = local.ssh_keys
+}
```
nazar-pc: Please configure your IDE to insert a single new line at the end of the files

resources/common.tf
```diff
@@ -55,9 +53,13 @@ data "digitalocean_ssh_key" "ved-key" {
   name = "Ved SSH Key"
 }
 
+# Sensitive variable. Contains private key for connecting to concrete Digital Ocean droplets
+variable "alexey2_do_private_key" { }
```
vedhavyas, line 57: why do we need to add this private key here?

resources/gemini-1b/provisioner.tf
```diff
@@ -28,8 +28,8 @@ resource "null_resource" "setup_nodes" {
     host = digitalocean_droplet.gemini-1b[count.index].ipv4_address
     user = "root"
     type = "ssh"
-    agent = true
-    agent_identity = var.ssh_identity
```
vedhavyas, line 32: why do you want to remove the identity?

resources/common.tf
ghost: I have done it for connecting cloud TF "null_resourse" to DO droplets.
Private key was saved in cloud TF secure/sensitive variable.

resources/gemini-1b/provisioner.tf
ghost: I haven't found another way to connect cloud terraform "null_resource" to droplets.
In case if you know any ways to do it - please share.

resources/common.tf
vedhavyas: I'm not super clear on the architecture. Can you add more details to understand what you thinking ?

My initial thought process was that we want to use a cloud state so that these private/sensitive variables never leave our workspaces. The added benefit with this is that everyone can use their application keys and its much easier for auditing the changes made while the state is kept updated so anyone can pick up with out sharing state explicitly.

PR #56, opened 2022-09-01
Gemini 2a infra

vedhavyas:
Add Gemini-2a infrastructure in DO
- Adds ~6~ 4 RPC nodes in APAC and NA each
- Adds 3 Bootstrap nodes in APAC and NA each

Note: I have removed gemini-1b as the infra is completely brought down outside this repo so syncing the state is not worth it


network-primitives/main.tf
```diff
@@ -0,0 +1 @@
+/Users/vedhavyas/Projects/infra/resources/common.tf
```
nazar-pc: Is this supposed to be committed?

nazar-pc (dismissed):
Quite involved and I wasn't trying to analyze everything, a bit outside of my scope.
Generally makes sense. `network-primitives` is something I'm not sure we need, seems complex and there is just one specific comment.

network-primitives/main.tf
vedhavyas: Ahh no. The soft links :(
Will take it out

vedhavyas:
Network primitives so that we can add another network with different config much much easily as most of the work is primitives. We can take it out if we don't use it for sure

nazar-pc (approved):

PR #57, opened 2022-09-14
Gemini-3a

vedhavyas:
This PR Adds Gemini-3a network and Removes DO droplets from Gemini-2a.

Gemini 3a updates include
- Docker args are updated with new cli args
- RPC nodes also run executor nodes for System and Payments domain

### DIFF
--- a/network-primitives/bootstrap_node_provisioner.tf
+++ b/network-primitives/bootstrap_node_provisioner.tf
@@ -22,15 +22,6 @@ resource "null_resource" "boostrap-node-keys" {
       NODE_PUBLIC_IPS = join(",", local.bootstrap_nodes_ip_v4)
     }
   }
-
-  # generate boostrap node keys
-  provisioner "local-exec" {
-    command     = "${var.path-to-scripts}/generate_node_keys.sh ${length(local.bootstrap_nodes_ip_v4)} ./dsn_bootstrap_node_keys.txt 50000"
-    interpreter = ["/bin/bash", "-c"]
-    environment = {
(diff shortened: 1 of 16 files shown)


network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -14,14 +14,23 @@ resource "null_resource" "boostrap-node-keys" {
...
+  # generate node keys
   provisioner "local-exec" {
     command     = "${var.path-to-scripts}/generate_node_keys.sh ${length(local.bootstrap_nodes_ip_v4)} ./bootstrap_node_keys.txt"
     interpreter = ["/bin/bash", "-c"]
     environment = {
       NODE_PUBLIC_IPS = join(",", local.bootstrap_nodes_ip_v4)
     }
   }
+
+  # generate boostrap node keys
```
nazar-pc: I don't think we use this currently, right? Or at least we're not using port 50000.

network-primitives/scripts/create_bootstrap_node_compose_file.sh
```diff
@@ -5,20 +5,33 @@ version: "3.7"
 
 volumes:
   archival_node_data: {}
+  boostrap_node_data: {}
 
 services:
+  dsn-bootstrap-node:
```
nazar-pc: I believe we decided to not deploy it separately for now.

nazar-pc (commented):
I think we should also specify explicitly block pruning. There must be `archive-canonical` (maybe for state too) that doesn't blow up in size from forking.

nazar-pc (approved):

PR #58, opened 2022-09-14
Add project for telemetry testing with single node same as for original telemetry

i1i1:


resources/common.tf
```diff
@@ -50,12 +50,16 @@ data "digitalocean_ssh_key" "serge-key" {
...
+data "digitalocean_ssh_key" "i1i1-key" {
+  name = "Ivan SSH Key"
+}
 
 # add ssh keys as single var
 locals {
   ssh_keys = [
     data.digitalocean_ssh_key.nazar-key.id,
     data.digitalocean_ssh_key.serge-key.id,
-    data.digitalocean_ssh_key.ved-key.id
```
vedhavyas: can you add your ssh key with in the project itself instead of common ? this will trigger changes for all the existing projects

vedhavyas (approved):

PR #59, opened 2022-10-16
Add Subsquid Archive: readme and `docker-compose.yml`

isSerge:


_docs/7_SubsquidArchive/docker-compose.yml
```diff
@@ -0,0 +1,83 @@
...
+      - "4444:3000"
+
+  node:
+    # Replace `snapshot-DATE` with latest release (like `snapshot-2022-apr-29`)
+    image: ghcr.io/subspace/node:snapshot-DATE
+    volumes:
+      # replace VOLUME_NAME with your volume name
+      - /mnt/VOLUME_NAME/node-data:/var/subspace:rw
+    ports:
+      - "0.0.0.0:9944:9944"
```
nazar-pc: Can we not expose RPC on public interface here?

_docs/7_SubsquidArchive/docker-compose.yml
```diff
@@ -0,0 +1,83 @@
...
+    command: [
+      "-e", "ws://node:9944",
+      "-c", "10",
+       "--prom-port", "9090",
+       "--out", "postgres://postgres:postgres@db:5432/squid-archive"
+    ]
+    environment:
+      NODE_TLS_REJECT_UNAUTHORIZED: 0
+    ports:
+      - "9090:9090" # prometheus port
```
nazar-pc: Not sure that here and in other places we actually want to expose these ports on public interfaces.
Maybe for some, but likely not at all.

achiurizo (approved):

nazar-pc (approved):
Looks good, though explorer should probably go through Nginx with HTTPS in production

isSerge:
> Looks good, though explorer should probably go through Nginx with HTTPS in production

I was not sure about this. If we use the archive for building a block explorer - it is still a long way for production. But if there is another use case for the archive, I could add Nginx with HTTPS right now.

PR #62, opened 2022-10-24
Add docker section

isSerge:


_docs/7_SubsquidArchive/README.md
```diff
@@ -2,6 +2,19 @@
...
 Subspace Network maintains [Subsquid Archive](https://docs.subsquid.io/archives/) in order to store raw chain data in a normalized way and expose it using GraphQL endpoint
 
+## Docker setup
+
+Install docker to run docker images:
+
+```bash
+sudo apt-get update
+sudo apt install -y apt-transport-https ca-certificates curl software-properties-common
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
nazar-pc: It is already available in stock repos and comparing to a few years back the version packaged isn't ancient. Can be replaced with just installation of `docker.io` and `docker-compose-plugin` (I think this is how modern version of compose is called, please double-check if you decide to use it).

nazar-pc (dismissed):
Looks fine, though might be simpler

_docs/7_SubsquidArchive/README.md
isSerge: I was using the script from https://github.com/subspace/infra/tree/main/_docs/5_SubspaceNetwork . Should we update there as well?

nazar-pc: I meant if you install `docker.io` and `docker-compose-plugin` from stock repos, you don't need to import keys, add new repos, download anything with `curl`, etc. It is literally this:
```bash
apt-get update
apt-get install -y docker.io docker-compose-plugin
```

There is nothing wrong with what is described in that link technically, but if you are in the mood feel free to update it.

nazar-pc (approved):

PR #64, opened 2022-11-03
Add GraphQL explorer to Nginx

isSerge:
- Adding GraphQL explorer to Nginx setup (now available at `https://archive.subspace.network/graphql`) 
- Moving Archive to `https://archive.subspace.network/api`
- Remove Certbot generated lines from the original Nginx config

### DIFF
--- a/_docs/7_SubsquidArchive/archive.subspace.network
+++ b/_docs/7_SubsquidArchive/archive.subspace.network
@@ -35,24 +35,4 @@ server {
         proxy_set_header Connection "upgrade";
         include /etc/nginx/cors-settings.conf;
     }
-
-    listen [::]:443 ssl ipv6only=on; # managed by Certbot
-    listen 443 ssl; # managed by Certbot
-    ssl_certificate /etc/letsencrypt/live/archive.subspace.network/fullchain.pem; # managed by Certbot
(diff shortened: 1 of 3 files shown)


achiurizo (dismissed):

_docs/7_SubsquidArchive/archive.subspace.network
```diff
@@ -1,14 +1,18 @@
...
-    root /var/www/arhive.subspace.network/html;
+    root /var/www/html;
     index index.html index.htm index.nginx-debian.html;
 
     server_name archive.subspace.network;
 
     location / {
-        try_files $uri $uri/ =404;
+        try_files $uri $uri/ @backend;
     }
```
nazar-pc: Why do we even have default location here?

isSerge: [GraphQL explorer UI](https://archive.subspace.network/graphql) requires some static files from `http://127.0.0.1:4444` - css, js...

nazar-pc: I see. It would be ideal to specify those subdirectories with static files explicitly though, just to be on a safe side and not serve `.env` file accidentally or whatnot. If there isn't anything like that in that location then it is not blocking.

nazar-pc (approved):

PR #68, opened 2022-12-18
Minor update Postgres config

isSerge:
Add `listen_addresses` into the Postgres config (required for PgHero to connect). The actual Postgres config in the prod currently uses a particular IP address instead of `'*'`


nazar-pc (dismissed):
Wouldn't it be better to put something invalid like `"REPLACE_IP"` so that it doesn't start with incorrect config?

nazar-pc (approved):

PR #69, opened 2022-12-19
add gemini-3b network

vedhavyas:


nazar-pc (commented):
Looks like this was created before we found that one more CLI argument is necessary for node.
Also please don't kill old network next time, leave at least one bootstrap and one RPC node before we announce and make sure no one needs it. We have sufficient number of machines idling, we can get rid of all full nodes if necessary.

nazar-pc (approved):

PR #70, opened 2022-12-20
USe Systemd to run subspace nodes and run netdata agent on each node

vedhavyas:
This PR brings following changes
- Run subspace nodes using Systemd so that we can use the hooks to add and remove firewall rules
- Run netdata on each node.

### DIFF
--- a/network-primitives/bootstrap_node_provisioner.tf
+++ b/network-primitives/bootstrap_node_provisioner.tf
@@ -70,7 +70,7 @@ resource "null_resource" "setup-bootstrap-nodes" {
       "sudo chmod +x /subspace/start_netdata_agent.sh",
       "sudo /subspace/start_netdata_agent.sh ${var.netdata_claim_token} ${var.netdata_claim_rooms} boostrap-node-${count.index}",
       "sudo iptables -I OUTPUT -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP",
-      "sudo iptables -I FORWARD -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP",
+      "sudo iptables -I FORWARD -o $(ip -o -4 route show to default | grep -E -o 'dev [^ ]*' | awk 'NR==1{print $2}') -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP",
     ]
   }
 
--- a/network-primitives/bootstrap_node_provisioner.tf
+++ b/network-primitives/bootstrap_node_provisioner.tf
@@ -63,14 +63,13 @@ resource "null_resource" "setup-bootstrap-nodes" {
   }
 
   # install docker and docker compose
+  # start netdata agent
   provisioner "remote-exec" {
     inline = [
       "sudo chmod +x /subspace/install_docker.sh",
       "sudo /subspace/install_docker.sh",
       "sudo chmod +x /subspace/start_netdata_agent.sh",
       "sudo /subspace/start_netdata_agent.sh ${var.netdata_claim_token} ${var.netdata_claim_rooms} boostrap-node-${count.index}",
-      "sudo iptables -I OUTPUT -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP",
-      "sudo iptables -I FORWARD -o $(ip -o -4 route show to default | grep -E -o 'dev [^ ]*' | awk 'NR==1{print $2}') -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP",
     ]
   }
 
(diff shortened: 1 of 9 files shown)


nazar-pc (commented):
Looks like they allow this to run under container and it'll be strongly preferred to do it that way IMO: https://learn.netdata.cloud/docs/agent/packaging/docker/

network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -69,6 +69,7 @@ resource "null_resource" "setup-bootstrap-nodes" {
       "sudo /subspace/install_docker.sh",
       "sudo chmod +x /subspace/start_netdata_agent.sh",
       "sudo /subspace/start_netdata_agent.sh ${var.netdata_claim_token} ${var.netdata_claim_rooms} boostrap-node-${count.index}"
+      "sudo iptables -I OUTPUT -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP",
```
nazar-pc: We do need to make this persistent though, right?

vedhavyas: Ideally yes, but I'm not sure how dynamic docker rules are as docker is also using them. Will check those and add the persistence.

vedhavyas:
> Looks like they allow this to run under container and it'll be strongly preferred to do it that way IMO: [learn.netdata.cloud/docs/agent/packaging/docker](https://learn.netdata.cloud/docs/agent/packaging/docker/)

True but In my personal experience, Its pain to deal with docker and netdata as we have to do bunch of extra steps to forward info from host to container for netdata to properly push container specific ones

At the moment, netdata is run as systemd service and felt easiet and more robout though.

I'm also looking at datadog and this will change anyway but will try to limit to docker like you suggested

network-primitives/bootstrap_node_provisioner.tf
vedhavyas: nvm, they are just docker specific networks. Iptables will be persisted

network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -119,6 +119,8 @@ resource "null_resource" "start-boostrap-nodes" {
       "sudo chmod +x /subspace/create_compose_file.sh",
       "sudo /subspace/create_compose_file.sh ${var.bootstrap-node-config.reserved-only} ${length(local.bootstrap_nodes_ip_v4)} ${count.index}",
       "docker compose -f /subspace/docker-compose.yml up -d --remove-orphans",
+      "sudo mkdir -p /etc/iptables/",
+      "sudo iptables-save > /etc/iptables/rules.v4",
```
nazar-pc: Hm... this will store Docker rules, I'd say we need to avoid it if at all possible, it will almost certainly become a source of problems at some point. Can you explicitly write just those that are necessary to block private networks?

vedhavyas: no, iptables-restore will overwrite the docker rules I believe. In this case, the is fine because we are not going to delete the docker networks. even if we did, the docker created its own iptable chains so i believe it should do a full replace within those chains.

and we are saving them after docker would have done all the updates to the chains

nazar-pc: I mean it will store whatever there is in iptables right now, which is not even guaranteed to be the same on all machines. If something changes in next Docker upgrade or whatever, it'll break. Do we need more than one line stored there? If not then `echo "..." > ` would be much more reliable.

network-primitives/bootstrap_node_provisioner.tf
nazar-pc: This doesn't seem to work, got another abuse message. Which doesn't make a lot of sense to me since we both fixed configuration on that machine and added these rules.

Should rules be added to FORWARD as well? Since container's network is private and packets are likely forwarded. I have added forward rule manually for now, though I don't like that I don't understand why this is happening on a fresh node.

Please run `sudo iptables -I FORWARD -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP` on all machines.

network-primitives/full_node_provisioner.tf
```diff
@@ -108,26 +107,46 @@ resource "null_resource" "start-full-nodes" {
...
+      # create docker compose file
       "sudo chmod +x /subspace/create_compose_file.sh",
+      "sudo chmod +x /usr/bin/subspace",
       "sudo /subspace/create_compose_file.sh ${var.bootstrap-node-config.reserved-only} ${length(local.full_node_ip_v4)} ${count.index} ${length(local.bootstrap_nodes_ip_v4)}",
-      "docker compose -f /subspace/docker-compose.yml up -d --remove-orphans",
-      "sudo mkdir -p /etc/iptables/",
-      "sudo iptables-save > /etc/iptables/rules.v4",
+
+      # start subspace node
+      "systemctl start subspace.service",
```
nazar-pc, line 149: I assume you've deleted `/etc/iptables` on those machines. Just checking.

nazar-pc (approved):
I see the idea is to run iptables rules before/after, but this doesn't protect from someone manually restarting containers, which is what I'd do if I had to debug something. Isn't there a way to make settings persistent even when Docker is involved?

network-primitives/full_node_provisioner.tf
vedhavyas: yes we did delete them.

vedhavyas:
> but this doesn't protect from someone manually restarting containers, which is what I'd do if I had to debug something. Isn't there a way to make settings persistent even when Docker is involved?

Restarting containers is fine but bringing `docker compose down` would break it. But we shouldn't be doing it now that systemd is involved. We update docker file as required and do `systemctl restart subspace` there by iptables are always upto date and docker containers are automatically updated as well

nazar-pc:
I see. This is really unfortunate that we need to put systemd on top of Docker, but I guess we'll have to live with it for now.

vedhavyas:
> I see. This is really unfortunate that we need to put systemd on top of Docker, but I guess we'll have to live with it for now.

true. docker team doesn't want to adds hooks. I think podman does but for some other day I guess

PR #71, opened 2023-01-05
Add `scan-start-value` parameter for `gateway` service

isSerge:
Currently processing blocks gets stuck when using the default value (100) on both Gemini-2a and Gemini-3b


vedhavyas (approved):

nazar-pc (approved):
Okay, it gets stuck, but why?

isSerge:
> Okay, it gets stuck, but why?

Because the batch size is too big, it takes longer to query DB, which leads to a timeout error. Decreasing batch size solves the problem

PR #72, opened 2023-01-05
remove, uninstall netdata agent and start datadog agent

vedhavyas:
This PR brings in following changes
- Remoe netdata agent running on the each node for node metrics
- Instead run datadog agent 

We made the move as discussed in previous discussions regarding using a datadog since 
  - we use it for logging as well 
  - prefer to reduce number of platforms
  - We are already paying for the datadog


nazar-pc (approved):
Please provide description with PR. Even I don't understand what has changed that you decided to switch from netdata to datadog all of a sudden. For others I suspect these changes even less meaningful.

vedhavyas:
Updated now. 
As moving it now, you suggested moving it datadog as we are using it anyway. I have intially went with netdata since it didn't require any administrative setup unlike Datadog

nazar-pc:
Got it, thanks!

PR #75, opened 2023-01-25
Devnet

vedhavyas:
This PR brings following changes:
- Devnet
- Ability to prune nodes
- Farmer node scripts to run farmer

~Note: Still need to update docker tags but good to review~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,9 @@ node_keys.txt
 /resources/gemini-3c/full_node_keys.txt
 /resources/gemini-3c/rpc_node_keys.txt
 /resources/gemini-3c/relayer_ids.txt
+/resources/devnet/relayer_ids.txt
+/resources/devnet/keystore/
+/resources/devnet/bootstrap_node_keys.txt
+/resources/devnet/farmer_node_keys.txt
+/resources/devnet/full_node_keys.txt
+/resources/devnet/rpc_node_keys.txt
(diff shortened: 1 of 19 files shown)


network-primitives/scripts/create_farmer_node_compose_file.sh
```diff
@@ -0,0 +1,94 @@
...
+
+  farmer:
+    depends_on:
+      archival-node:
+        condition: service_healthy
+    image: ghcr.io/\${NODE_ORG}/farmer:\${NODE_TAG}
+    volumes:
+      - farmer_data:/var/subspace:rw
+    restart: unless-stopped
+    command: [
```
nazar-pc: Farmer should also expose DSN port or you didn't want that?

.gitignore
```diff
@@ -18,3 +18,9 @@ node_keys.txt
 /resources/gemini-3c/full_node_keys.txt
 /resources/gemini-3c/rpc_node_keys.txt
 /resources/gemini-3c/relayer_ids.txt
+/resources/devnet/relayer_ids.txt
```
nazar-pc, line 21: What is relayer doing here?

nazar-pc (dismissed):

.gitignore
vedhavyas: this hold the relayer ids used to run the relayers for domains

network-primitives/scripts/create_farmer_node_compose_file.sh
vedhavyas: I was not aware farmer can. Thought not sure if we want to. Lets leave it as is for now

nazar-pc: We have an example with ports in the farming docs with Docker in monorepo

nazar-pc (approved):

PR #76, opened 2023-01-27
start farmer node for gemini-3c

vedhavyas:
This PR starts a farmer for Gemini-3c


nazar-pc (commented):
I'm really not sure we want to run farmer for Gemini 3c, it is devs and community who should run them

network-primitives/variables.tf
```diff
@@ -13,6 +13,12 @@ variable "datadog_api_key" {
   type        = string
 }
 
+variable "piece_cache_size" {
+  description = "Datadog API Key"
```
nazar-pc: Copy-pasted comment

resources/gemini-3c/main.tf
```diff
@@ -4,52 +4,53 @@ module "gemini-3c" {
...
     docker-org             = "subspace"
-    docker-tag             = "gemini-3c-2023-jan-18"
+    docker-tag             = "gemini-3c-2023-jan-27"
     reserved-only          = false
     plot-size              = "10G"
     reward-address         = var.farmer-reward-address
     force-block-production = false
     prune                  = false
   }
+  piece_cache_size     = "500KiB"
```
nazar-pc: Why 500KiB? We want something like 10G at least instead. I was talking about 500k pieces, not bytes.

vedhavyas:
> I'm really not sure we want to run farmer for Gemini 3c, it is devs and community who should run them

Fair but lets keep one running so that we don't let the chain stall.

nazar-pc (approved):

PR #82, opened 2023-03-13
Subspace DNS resources

### LINKED ISSUE #88: Migrate DNS records to infra
Migrate all the DNS records that are created manually on Cloudflare to infra repo.

vedhavyas:
This PR aims to track all the DNS records that are not created/tracked in this Repo.  

Records are imported to TF state. Updated TTLs where applicable to 3600.

Closes #88


resources/dns/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How to add a new DNS record?
+A new DNS record can be added to records.tf. Here is an example
+
+```
+resource "cloudflare_record" "name_of_the_resource" {
+  name    = "name_of_the_resource"
+  comment = "comment explaining the record existence"
+  proxied = true/false
+  ttl     = 1
```
nazar-pc: Let's correct example to use sane value

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+}
+
+resource "cloudflare_record" "mailserver_4" {
+  name     = "subspace.network"
+  comment  = "MX record pointing to our preferred mailserver"
+  priority = 10
+  proxied  = false
+  ttl      = 3600
+  type     = "MX"
+  value    = "alt3.aspmx.l.google.com"
```
nazar-pc: 1,2,4,3. Why not 1,2,3,4? :slightly_smiling_face:

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+  zone_id  = data.cloudflare_zone.cloudflare_zone.id
+}
+
+resource "cloudflare_record" "mailserver_default_domain_key" {
+  name    = "default._domainkey"
+  comment = "DomainKeys authentication for mail"
+  proxied = false
+  ttl     = 3600
+  type    = "TXT"
+  value   = "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA7zwrsOLB+eJ9SG1t7+OwOT2BuacTImFHozl8I/ mypg7rhtp4i1NpkSnjjDC3FdXXiUhsHTvAUvg5nMtGp3nCwwQYna0C8Jo7dbt3+NUVLj9KCBBBegxPS/WoJghPSbiKq4T/SBdM0K ShrVn7C/1blWA+N4XxOwmVtELV8POMwRYCzlrxCi3kdjbRY+4gXYKmcc7MSRi5ubyR7P/+K1/CkLbJa1SbxdS6/zMRIzPH/6vOR6 be1Qkw5PFsYu0gYbz3QDqYxaUTS3euWSPE3uLnPQEgAryX3SlKQB8uyjNbxF86ukslwwe9Q5cCZ1UPsl/89qcBRirnJbfoGlwb5j rX3QIDAQAB;"
```
nazar-pc: I'm 95% positive this is broken, DMIK public key shouldn't contain whitespaces. They are required in Google Cloud, but they don't work in CloudFlare, resulting in emails being unverifiable and having higher change of getting into SPAM.

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+  zone_id = data.cloudflare_zone.cloudflare_zone.id
+}
+
+resource "cloudflare_record" "mailserver_dmarc" {
+  name    = "_dmarc"
+  comment = "DMARC for communications"
+  proxied = false
+  ttl     = 3600
+  type    = "TXT"
+  value   = "v=DMARC1; p=reject; pct=100; rua=mailto:g5oqy2di@ag.dmarcian.com, mailto:admin@subspace.network; ruf=mailto:admin@subspace.network"
```
nazar-pc: There isn't much benefit in sending this to admin if we're using dmarcian, does anyone actually read those reports? @ImmaZoni?

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+  comment  = "MX record pointing to our preferred mailserver"
+  priority = 10
+  proxied  = false
+  ttl      = 3600
+  type     = "MX"
+  value    = "alt3.aspmx.l.google.com"
+  zone_id  = data.cloudflare_zone.cloudflare_zone.id
+}
+
+resource "cloudflare_record" "mailserver_default_domain_key" {
```
nazar-pc: I'd use `dkim` instead of `domain_key`

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+
+resource "cloudflare_record" "mailserver_google_domain_key" {
+  name    = "google._domainkey"
+  comment = "DomainKeys authentication for mail specific to Google mailserver"
+  proxied = false
+  ttl     = 3600
+  type    = "TXT"
+  value   = "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyqbz3f5O6pOl9LK3zOmKkczcK9aCdB34tSzDkA ZRD0g+OtBhlyWyzNeQGsrVianIfM9xyfZ7MVJK7sB/VGKIFIe5glb/Lh9tf/kLCRbkOnaafiXP5tOk4DC+mBpHOjyT0GgX5x4hxg oLmeJOrRLSu1niPQY/VqMtNsYa9+gAIo0YZnZeNi2w3FjWaslm5F7uI6mISdH3HZchqhzx2E6Ct+VJhLM0Ir+6v3qV5ylArtJzc+ PoTGdk65n5oq+Ioj9oTK4VmJunZ6jaU2Vimo+2TTAhBtRWQCa8Xy15/3kVZXiKJ4ddtPDnzSEval4er4SksIhtVYaxkeanU7pGbK w1SQIDAQAB"
+  zone_id = data.cloudflare_zone.cloudflare_zone.id
+}
```
nazar-pc: This is the same as above (except `;` is added unnecessarily at the end of the above record). I don't think this should be necessary, seems accidental, @ImmaZoni do you know who set this up?

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+  value   = "162.159.153.4"
+  zone_id = data.cloudflare_zone.cloudflare_zone.id
+}
+
+// subspace docs
+resource "cloudflare_record" "docs" {
+  name    = "docs"
+  comment = "Redirected docs to Github"
+  proxied = true
+  ttl     = 1
```
nazar-pc, line 31: TTL here and in some records below is still 1 second. Most DNS servers will ignore it anyway, why not specifying something more reasonable?

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+  value   = "subspace.github.io"
+  zone_id = data.cloudflare_zone.cloudflare_zone.id
+}
+
+resource "cloudflare_record" "terraform_managed_resource_964bf73d0853cefc64803baef62167a5" {
+  name    = "ws"
+  proxied = true
+  ttl     = 1
+  type    = "A"
+  value   = "65.108.232.54"
```
nazar-pc, line 76: Proxied IP addresses should ideally be private, so they are not reachable directly (though better yet would be to use CloudFlare's tunnels at this point to hide completely, but that is for another time).

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+}
+
+resource "cloudflare_record" "terraform_managed_resource_61eda5c4af93ca9d0f50e2fe1b56ae6d" {
+  name    = "subspace.network"
+  proxied = false
+  ttl     = 3600
+  type    = "TXT"
+  value   = "google-site-verification=Fxi1PlaR_c4E2l34q5G5QoiMPc7PeRqY_sLpD4XiAxg"
+  zone_id = data.cloudflare_zone.cloudflare_zone.id
+}
```
nazar-pc, line 276: I'd add TODOs to check if these are necessary, I'm quite certain it is not.

resources/dns/records.tf
```diff
@@ -0,0 +1,379 @@
...
+  name    = "zh"
+  proxied = false
+  ttl     = 3600
+  type    = "CNAME"
+  value   = "websites.weglot.com"
+  zone_id = data.cloudflare_zone.cloudflare_zone.id
+}
+
+resource "cloudflare_record" "terraform_managed_resource_d5b9ae2a99064d91811927a96fda5d57" {
+  name    = "_smtp._tls"
```
nazar-pc, line 243: This is a strange record, not sure if it is necessary either.

nazar-pc (commented):
Thanks for working on this!
I think overall it makes sense, though splitting into multiple files (like moving all email-related stuff into separate file, all website-related into another) would be helpful for maintenance.

resources/dns/records.tf
vedhavyas: for proxied records, cloudflare enforce to be 1

vedhavyas:
As for DKIM records having spaces. That is totally fine, I have already validated them. As for whther the keys are valid or not is another question.

nazar-pc:
If you didn't check whether signature verifies properly by receiving side, it is not totally fine.

But I did check and they appear to be working correctly (I validated that signature is correct), though I don't quite understand why, see https://github.com/Mailu/Mailu/issues/2618 where I saw a similar issue where public key was sliced and that caused issues :man_shrugging:

vedhavyas:
hey @ImmaZoni @nazar-pc is there any concerns we have to handle here or we good to make it ready for review ?

nazar-pc (approved):
I don't have major concerns beyond what I already shared.
While it is not perfect, as initial dump to move things from manually edited Cloudflare should be fine.

PR #84, opened 2023-03-15
Add DSN connection limit parameters.

shamil-gadelshin:
This PR changes docker compose yaml creation scripts for bootstrap and full nodes:
--dsn-max-in-connections
--dsn-max-out-connections 

Should be applied only after https://github.com/subspace/subspace/pull/1250 is merged.


vedhavyas (approved):
We okay not adding them to RPC and other node?

shamil-gadelshin:
> We okay not adding them to RPC and other node?

No, I don't think we need it now.

PR #93, opened 2023-03-20
add resources directory into a terraform folder

DaMandal0rian:
Wanted to move terraform IaC into it's own folder since I plan to use other infra folders such as ansible and a docker folder for dockerfiles/compose etc..


nazar-pc (commented):
Are you sure we want to have all of those in the same repo?
My assumption was that this repo is about infrastructure, which is almost 100% terraform in our case.

Things related to deployment are probably better kept separate from this.

As to Dockerfiles, more context would be helpful. Most of the Dockerfiles live in the repo with the software they build.

DaMandal0rian:
@nazar-pc Ok, usually infra related stuff are all bundled into "infra" repo, which is usually IaC, Configuration management (i.e ansible), monitoring, logging etc... But I'm open to use your structuring.
As for dockerfiles/compose, this is usually from infra related stuff like nginx, dbs, load-balancers etc..

nazar-pc:
I think ansible for initial deployment of machines is fine, not sure what exactly you mean when talking about monitoring and logging, those seems like software deployment on top of infrastructure rather that infrastructure itself.

DaMandal0rian:
My thinking was that the infra repo was everything related to infra with a structure like:

├── ansible
├── configs
├── packer
├── kubernetes
├── load-balancers
├── scripts
└── terraform

PR #105, opened 2023-04-02
fix broken prune command.

vedhavyas:
Seems like system prune will not delete any named volumes after the recent update due to change in default. More on this here - https://github.com/docker/cli/issues/4028. For now, use the docker volume prune instead

### DIFF
--- a/network-primitives/scripts/prune_docker_system.sh
+++ b/network-primitives/scripts/prune_docker_system.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 
 systemctl stop subspace.service
+# disable the service
+systemctl disable subspace.service
+# delete the service file
+rm  /etc/systemd/system/subspace.service
 # update docker restart policy to no so that containers wont start.
 docker ps -aq | xargs docker update --restart=no
 # stop all the containers
--- a/network-primitives/scripts/prune_docker_system.sh
+++ b/network-primitives/scripts/prune_docker_system.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+systemctl stop subspace.service
+# update docker restart policy to no so that containers wont start.
+docker ps -aq | xargs docker update --restart=no
+# stop all the containers
+docker ps -aq | xargs docker stop
+# prune everything docker related
+docker system prune -a -f --volumes
+# seems like system prune is broken with docker 23.
+# so we need to prune volumes manually. Remove the following when issue is fixed.
+# TODO: https://github.com/docker/cli/issues/4028
+docker volume prune -f --filter all=1
--- a/network-primitives/bootstrap_node_provisioner.tf
+++ b/network-primitives/bootstrap_node_provisioner.tf
@@ -83,11 +83,16 @@ resource "null_resource" "prune-bootstrap-nodes" {
     timeout        = "10s"
   }
 
+  provisioner "file" {
+    source      = "${var.path-to-scripts}/prune_docker_system.sh"
+    destination = "/tmp/prune_docker_system.sh"
+  }
+
   # prune network
   provisioner "remote-exec" {
     inline = [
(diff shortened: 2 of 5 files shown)


nazar-pc (commented):
Why are we using `docker system` at all? I thought we just have compose on one machine, in which case `docker compose` is more appropriate.

vedhavyas:
Because I was going more defensive in cleaning all the docker related object instead of limiting to just the compose project since we are not using these machines for anything else. Do you have any concerns ?

nazar-pc:
I don't see us using `docker system` at all, if we do we can add it to the list, otherwise just raises unnecessary questions, for me at least

vedhavyas:
Hmmm I don't see other way to clean up the objects that don't fall within the scope the project using docker compose. Is there another way to do that ?

nazar-pc:
What are you even trying to do here? Maybe you want to stop Docker service and wipe `/var/lib/docker` instead?

vedhavyas:
Why ?
This does the exact same thing.

> What are you even trying to do here? 

Trying to remove unrelated docker objects that may have been there .

Not sure what is your concern here tbh

nazar-pc:
You're trying to run a bunch of commands to clean up everything Docker-related and still missing a bunch of stuff: images, networks, potentially something else.

Just stop the service, delete `/var/lib/docker` and start again.

vedhavyas:
This system prune does exactly what you expect.
https://docs.docker.com/engine/reference/commandline/system_prune/

I would rather not do a `rm` when the the cli does it better 😉

nazar-pc:
Didn't know about that command. However even then current implementation doesn't work properly on containers that have `restart: always`, they'll start right after you `docker stop` them.

Also if you add workaround for a bug, make sure to add TODO with explanation and link to upstream issue so we know when to drop it. Seeing `docker system prune` followed by `docker volume prune` makes no sense if you read it in isolation.

nazar-pc (dismissed):
:man_shrugging:

network-primitives/scripts/prune_docker_system.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+systemctl stop subspace.service
```
nazar-pc, line 3: BTW, shouldn't we delete the service afterwards. It will try to start after reboot.

nazar-pc (dismissed):

DaMandal0rian (dismissed):

network-primitives/scripts/prune_docker_system.sh
```diff
@@ -0,0 +1,13 @@
...
+# update docker restart policy to no so that containers wont start.
+docker ps -aq | xargs docker update --restart=no
+# stop all the containers
+docker ps -aq | xargs docker stop
+# prune everything docker related
+docker system prune -a -f --volumes
+# seems like system prune is broken with docker 23.
+# so we need to prune volumes manually. Remove the following when issue is fixed.
+# TODO: https://github.com/docker/cli/issues/4028
+docker volume prune -f --filter all=1
```
DaMandal0rian, line 17: For line 9 & 13 can't this be achieved with
`docker volume ls -q | xargs docker volume rm -f` ?

vedhavyas: No, it wont prune other docker related object. Just volumes

DaMandal0rian: sorry I should have added the following for all resources
`docker system prune --all -f && docker volume ls -q | xargs docker volume rm -f `

DaMandal0rian (approved):

network-primitives/scripts/prune_docker_system.sh
vedhavyas: Either is fine and give you the same results. I'm keep it as is for now unless you have a strong opinion not to 😅

PR #111, opened 2023-04-13
Update DSN ports and upgrade to april release

shamil-gadelshin:
This PR updates DSN ports from 50000->50001 and 30433->30434 for several scripts.

### DIFF
--- a/network-primitives/scripts/create_bootstrap_node_compose_file.sh
+++ b/network-primitives/scripts/create_bootstrap_node_compose_file.sh
@@ -36,7 +36,7 @@ services:
     restart: unless-stopped
     ports:
       - "30333:30333"
-      - "30434:30434"
+      - "30434:30433"
     command: [
       "--chain", \$NETWORK_NAME,
       "--base-path", "/var/subspace",
--- a/network-primitives/scripts/create_bootstrap_node_compose_file.sh
+++ b/network-primitives/scripts/create_bootstrap_node_compose_file.sh
@@ -27,7 +27,11 @@ services:
       - RUST_LOG=info
     ports:
       - "50001:50001"
-    command: ["start", \$DSN_NODE_KEY, "/ip4/0.0.0.0/tcp/50001" ]
+    command: [
+      "start",
+       \$DSN_NODE_KEY, "/ip4/0.0.0.0/tcp/50001",
+        "--protocol-version", \$GENESIS_HASH
+      ]
 
   archival-node:
     image: ghcr.io/\${NODE_ORG}/node:\${NODE_TAG}
--- a/network-primitives/scripts/create_bootstrap_node_compose_file.sh
+++ b/network-primitives/scripts/create_bootstrap_node_compose_file.sh
@@ -26,8 +26,8 @@ services:
     environment:
       - RUST_LOG=info
     ports:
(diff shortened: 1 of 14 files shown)


nazar-pc (commented):
Well, since these are primitives, they will likely break devnet and whatever other networks we have by changing their ports as well, so I'm not 100% sure about this one

shamil-gadelshin:
Since we're not sure about this PR - I'm converting it to the draft until we cancel it. 
Also, keep in mind if you're going to merge it after all that we need to add the "--protocol-version" CLI argument for bootstrap-node with the value equals to the genesis hash of gemini-3d

network-primitives/scripts/create_bootstrap_node_compose_file.sh
```diff
@@ -36,7 +36,7 @@ services:
     restart: unless-stopped
     ports:
       - "30333:30333"
-      - "30433:30433"
```
vedhavyas, line 39: Are you sure about the this change? Looking at this PR - https://github.com/subspace/subspace/pull/1371/files
I think what you want is this instead
`30434:30433`

vedhavyas (commented):
network-primitives should expose the dsn listen port with default as a variable through the config so that we can customise it just for gemini-3d. I can push those changes if you are not comfortable with terraform.

network-primitives/scripts/create_bootstrap_node_compose_file.sh
shamil-gadelshin: Indeed. I wasn't going to change the default ports. Will fix it now.

shamil-gadelshin:
I fixed the container ports back to the default (30433) and added `--protocol-version` (GENESIS_HASH) parameter to the bootstrap-application  - we need it after https://github.com/subspace/subspace/pull/1377 . However, I'm not aware of how do we set the environment variables.

vedhavyas:
@shamil-gadelshin all good. I will take over from here to make necessary changes and get this in. Thanks!

network-primitives/scripts/create_bootstrap_node_compose_file.sh
```diff
@@ -26,8 +26,13 @@ services:
     environment:
       - RUST_LOG=info
     ports:
-      - "50000:50000"
-    command: ["start", \$DSN_NODE_KEY, "/ip4/0.0.0.0/tcp/50000" ]
+      - "50001:50001"
```
nazar-pc: Shouldn't this be configurable as well?

network-primitives/variables.tf
```diff
@@ -31,6 +31,7 @@ variable "full-node-config" {
     docker-tag          = string
     reserved-only       = bool
     prune               = bool
+    node-dsn-port       = string
```
nazar-pc: Is there a `number` type? Seems more appropriate for the port.

nazar-pc (dismissed):

network-primitives/scripts/create_bootstrap_node_compose_file.sh
vedhavyas: actually, yes. Good catch

nazar-pc (approved):

PR #112, opened 2023-04-14
upgrade gemini-3d latest april release

DaMandal0rian:
This PR upgrades the Geimini-3d nodes to latest release apr-14.


resources/gemini-3d/main.tf
```diff
@@ -4,34 +4,34 @@ module "gemini-3d" {
   network-name    = "gemini-3d"
   bootstrap-node-config = {
     droplet_size        = var.droplet-size
-    deployment-version  = 6
+    deployment-version  = 7
     regions             = []
-    nodes-per-region    = 0
+    nodes-per-region    = 1
```
vedhavyas, line 9: we do not needs to update this since we dont use Digital ocean for this. If we want to deploy to digital ocean, then you give one or many regiions and nodes-per-region will create # nodes per region. Right now since regions are empty, it wont account for any nodes in DO

PR #124, opened 2023-05-05
devnet changes

DaMandal0rian:
This PR enables domains on all components and also sets a domain_id attribute.


network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -174,7 +174,7 @@ resource "null_resource" "start-boostrap-nodes" {
       # create docker compose file
       "sudo chmod +x /subspace/create_compose_file.sh",
       "sudo chmod +x /usr/bin/subspace",
-      "sudo /subspace/create_compose_file.sh ${var.bootstrap-node-config.reserved-only} ${length(local.bootstrap_nodes_ip_v4)} ${count.index}",
```
vedhavyas: `domain_id` is not required for all just rpc since that is where we run domains

vedhavyas (commented):
We just need to update RPC nodes. rest can be left as is. 

Also need to update other resources apart from devnet. They should not give any terrform changes

DaMandal0rian:
> We just need to update RPC nodes. rest can be left as is.
> 
> Also need to update other resources apart from devnet. They should not give any terrform changes

@vedhavyas this branch is for devnet only, will open a PR for the gemini-3d terraform

network-primitives/variables.tf
```diff
@@ -32,6 +32,7 @@ variable "full-node-config" {
     reserved-only       = bool
     prune               = bool
     node-dsn-port       = number
+    domain-id           = number
```
vedhavyas: forgot to remove these ?

resources/devnet/variables.tf
```diff
@@ -30,3 +30,9 @@ variable "farmer-reward-address" {
   description = "Farmer's reward address"
   type        = string
 }
+
+variable "domain_id" {
+  description = "Domain ID"
+  type        = number
+  default     = 1
+}
```
vedhavyas: new line pls

vedhavyas (commented):
Make sense. Some nits and there are some forgotten changes that are not required

network-primitives/variables.tf
DaMandal0rian: thanks, good catch!

DaMandal0rian: i actually removed them, and my git work tree is clean. Need to check why it wasn't staged. :/

PR #130, opened 2023-05-22
changes network primitives

DaMandal0rian:
- enable decoupled domain support
- update gemini-3d
- update devnet

### DIFF
--- a/network-primitives/scripts/create_domain_node_compose_file.sh
+++ b/network-primitives/scripts/create_domain_node_compose_file.sh
@@ -45,14 +45,14 @@ services:
     labels:
       caddy_0: \${DOMAIN_PREFIX}-\${NODE_ID}.system.\${NETWORK_NAME}.subspace.network
       caddy_0.handle_path_0: /http
-      caddy_0.handle_path_0.reverse_proxy: "{{upstreams 9933}}"
+      caddy_0.handle_path_0.reverse_proxy: "{{upstreams 8933}}"
       caddy_0.handle_path_1: /ws
-      caddy_0.handle_path_1.reverse_proxy: "{{upstreams 9944}}"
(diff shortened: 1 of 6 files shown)


network-primitives/scripts/create_domain_node_compose_file.sh
```diff
@@ -112,7 +113,7 @@ if [ ${enable_domains} == true ]; then
       echo '      "--validator",'
       echo '      "--state-pruning", "archive",'
       echo '      "--blocks-pruning", "archive",'
-      echo '      "--base-path", "/var/subspace/system_domain",'
```
vedhavyas: why comment them ?

DaMandal0rian: the docker compose manifest cant run because it has multiple "base-path" declarations. Just commented for now in case we need to switch the paths used per domain. It's currently set 3 times.

vedhavyas: that was intentional since each one corresponds to Primary, System and Domains.
If not, the base path will be same for all the three and will corrupt the data

vedhavyas (commented):
looks good overall

network-primitives/scripts/create_domain_node_compose_file.sh
DaMandal0rian: Understand, I gathered that too as intentional, but the --base-path flag can't be set multiple times. Either different docker service per domain be created, or a way for the node to write to different folders per domain under the root base-path.

DaMandal0rian: This was a bug with quotes and escaping, so comment above irrelevant.

vedhavyas (dismissed):

vedhavyas (approved):

Issue #131, opened 2023-05-23
Create vault instance for key syncing


DaMandal0rian:
- [x] The first PR #132  creates the vault instance on AWS infrastructure

PR #132, opened 2023-05-23
Vault Infra

DaMandal0rian:
This PR creates a hashicorp vault single cluster instance on AWS using terraform.
 - Backend storage S3 (encrypted / private acl)
 - KMS key to manage vault seal
 - IAM user to launch the ec2 instance with S3 access enabled.

This is required for the infrastructure to automate the process of handling keys and network deployments with CI and gitops.

Related Issue #131

### DIFF
--- a/explorer/.gitignore
+++ b/explorer/.gitignore
@@ -1,0 +1,6 @@
+.terraform*
+*.tfstate*
+*.tfplan
+.DS_Store
+backend_config.hcl
+*.tfvars
--- a/explorer/.gitignore
+++ b/explorer/.gitignore
@@ -1,6 +1,0 @@
-.terraform*
-*.tfstate*
-*.tfplan
-.DS_Store
-backend_config.hcl
-*.tfvars
--- a/vault/terraform/main.tf
+++ b/vault/terraform/main.tf
@@ -139,7 +139,7 @@ data "local_file" "private_key" {
 }
 
 resource "aws_instance" "vault" {
-  ami                         = data.aws_ami.ubuntu_x86.id # Update with the desired Vault AMI ID
+  ami                         = data.aws_ami.ubuntu_amd64.id # Update with the desired Vault AMI ID
   instance_type               = var.vault_instance_type    # Update with the desired instance type
   key_name                    = var.ssh_key_name           # Update with your SSH key pair
   subnet_id                   = aws_subnet.vault_subnet.id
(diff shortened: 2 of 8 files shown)


vault/terraform/main.tf
```diff
@@ -0,0 +1,206 @@
...
+  vpc_security_group_ids = [aws_security_group.vault_sg.id]
+
+  tags = {
+    Name = "vault-server"
+  }
+
+  provisioner "remote-exec" {
+  inline = [
+    "export DEBIAN_FRONTEND=noninteractive",
+    "sudo apt update -y",
```
nazar-pc: Just wondering, do all of these actually need sudo? I'd expect provisioner to run under root already

vault/terraform/main.tf
```diff
@@ -0,0 +1,206 @@
...
+    "sudo apt upgrade -y",
+    "sudo apt install git curl wget gnupg openssl make build-essential jq net-tools unzip nginx certbot python3-certbot-nginx -y",
```
nazar-pc: Are all of these actually necessary? I don't see `make`, `jq` and others used anywhere.

explorer/.gitignore
```diff
@@ -0,0 +1,6 @@
+.terraform*
+*.tfstate*
+*.tfplan
+.DS_Store
+backend_config.hcl
+*.tfvars
```
nazar-pc: This seems to be committed by accident

vault/terraform/main.tf
DaMandal0rian: yes, sudo is needed for ubuntu user.

explorer/.gitignore
DaMandal0rian: good catch thanks.

vault/terraform/main.tf
DaMandal0rian: I could remove some of them and install later if needed.

DaMandal0rian (commented):
will make the minor changes

DaMandal0rian (commented):
will make the minor changes

vault/terraform/main.tf
nazar-pc: We should only install packages that are required for a specific machine, not the same set of packages on all machines regardless of the role. Latest version of this PR still installs with `wget` ans `curl`. Do we really actually need both though?

vedhavyas (commented):
Looks good overall

vault/terraform/main.tf
DaMandal0rian: done! removed wget.

nazar-pc (approved):

Issue #122, opened 2023-05-01
Migrate from DO to AWS


0xjeremyfrank:
- [x] explorers
- [x] gemini networks
- [x] telemetry

DaMandal0rian:
All infra templates have been setup and tested. Next is to actually migrate everything and shutdown old servers.

PR #134, opened 2023-05-26
AWS migration devnet and gemini

DaMandal0rian:
The PR migrates our devnet network and gemini infra from DO to AWS
- this is related to issue #122 
- start refactoring for #135


vedhavyas (changes requested):
This is not the way to go. IIRC, we wanted to move away from current network primitives to something like ansible i believe. Maybe @nazar-pc could chime in here regarding that.

Having said that, if we still want to use network primitives there is no reason to copy and create a new set of same scripts. The network primitive's scripts and a good chunk of terraform scripts like provisioner are independent of the cloud. So there is no difference between Digital ocean or Heztner there once instances are created since we just need SSH access. If we want to integrate AWS and deprecate DO, then add the required variables exposed through the main module like DO for AWS.

DaMandal0rian:
> This is not the way to go. IIRC, we wanted to move away from current network primitives to something like ansible i believe. Maybe @nazar-pc could chime in here regarding that.
> 
> Having said that, if we still want to use network primitives there is no reason to copy and create a new set of same scripts. The network primitive's scripts and a good chunk of terraform scripts like provisioner are independent of the cloud. So there is no difference between Digital ocean or Heztner there once instances are created since we just need SSH access. If we want to integrate AWS and deprecate DO, then add the required variables exposed through the main module like DO for AWS.

1. Please see https://github.com/subspace/infra/issues/135 which discusses this. The provisioners are not ideal with AWS and as simple as you may think it is with just using SSH it is not because the current deployment is tied to root user. The created user on AWS is `ubuntu`. Hetzner and DO were using root user for deployments. Not sure if your concern is a blocker, a new PR will be done to remove those scripts and provisioners once #135 is ready.

vedhavyas:
> The created user on AWS is ubuntu. Hetzner and DO were using root user for deployments.

Then we should make those scripts not assume root. 

> a new PR will be done to remove those scripts and provisioners once

No, why cant we use the exisiting scripts already ?

>  The provisioners are not ideal with

Sure but this PR still uses those provisioners. I dont see any ansible or other deployment tools here 🤔 
Am I missing something?

> Please see https://github.com/subspace/infra/issues/135 which discusses this.

This doesn;t give me any extra information apart from `root` which we can adjust the exisiting scripts.


> Not sure if your concern is a blocker,

My main blocker is the duplicated scripts that exactly does the same thing

DaMandal0rian:
> > The created user on AWS is ubuntu. Hetzner and DO were using root user for deployments.
> 
> Then we should make those scripts not assume root.
> 

Right, so wouldn't that require a rewrite? Doesn't systemd and docker need a sudo user?

> > a new PR will be done to remove those scripts and provisioners once
> 
> No, why cant we use the exisiting scripts already ?
> 

Because they can't work as currently is, welcome to try on AWS.

> > The provisioners are not ideal with
> 
> Sure but this PR still uses those provisioners. I dont see any ansible or other deployment tools here thinking Am I missing something?
>

No you're not but why include deployment and ansible with infrastructure?
 
> > Please see #135 which discusses this.
> 
> This doesn;t give me any extra information apart from `root` which we can adjust the exisiting scripts.
>
Not sure if a thesis is required to discuss why a deployment works with root user and wouldn't with ubuntu or any other user with existing scripts. 

Yes with sudo right? Working on that as a temporary solution already for deployment but it's not ideal and thats why I have #135 . I will upload the commit though.
 
> > Not sure if your concern is a blocker,
> 
> My main blocker is the duplicated scripts that exactly does the same thing

Actually I've shaved out alot of the scripts and changed them, is that alright i include them once tested and working?

Lastly, your last reply is very contradictive, do you you want to use the same scripts, change scripts to not assume root, use some other deployment tool?

vedhavyas:
> Lastly, your last reply is very contradictive, do you you want to use the same scripts, change scripts to not assume root, use some other deployment tool?

Well, I would like for us to use the same scripts if we are taking that route. But you and @nazar-pc did not like that approach. If we going to use the ansible or other then we should not use these scripts and the deployment is done using the chosen tool completely. These scripts generate docker compose files and the chose tool should do it instead of these scripts.

But considering the changes you have done, seems like we are going to use the same approach we currently have, then the scripts needs to be updated to work AWS instead of creating a new set of scripts and rewrite is not necesary. Systemd can run without root as well.

> Not sure if a thesis is required to discuss why a deployment works with root user and wouldn't with ubuntu or any other user with existing scripts.

Again sarcasm doesn't help anyone. I'm merely suggesting an approach since you asked my review. I'm happy to sit out on this and let you and @nazar-pc handle this completely :)

aws/devnet/main.tf
```diff
@@ -0,0 +1,97 @@
...
+    additional-node-ips = var.domain_node_ips
+    docker-org          = "subspace"
+    docker-tag          = "gemini-3d-2023-may-23"
+    domain-prefix       = "eu"
+    reserved-only       = false
+    prune               = false
+    node-dsn-port       = 30434
+    enable-domains      = true
+    domain-id           = var.domain_id
+    domain-labels       = var.domain_labels
```
nazar-pc: Just out of curiosity, why are some things using `-` as separator and others `_`? Is this some kind of naming convention? Looks a bit inconsistent.

aws/devnet/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+  description = "Domain ID"
+  type        = list(number)
+  default     = [1, 2, 3]
+}
+
+variable "domain_labels" {
+  description = "Tag of the domain to run"
+  type        = list(string)
+  default     = ["evm", "payments"]
+}
```
nazar-pc: A bit confusing that one has 2 elements and the other has 3

aws/devnet/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+
+
+variable "bootstrap_node_ips" {
+  description = "IP of boostrap node"
+  type        = list(string)
+  default     = [""]
```
nazar-pc: Do you need this default value? It makes no sense as an IP address, it'd better fail and request user to provide an IP instead of silently using invalid value.

aws/devnet/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+variable "azs" {
+  type        = list(string)
+  description = "Availability Zones"
+  default     = ["us-east-1a", "us-east-1b"]
+}
+
+variable "instance_count" {
+  type    = number
+  default = 1
+}
```
nazar-pc, line 37: Is this per zone or global?

aws/devnet/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+variable "aws_region" {
+  description = "aws region"
+  type        = list(string)
+  default     = ["us-east-1"]
+}
+
+variable "public_subnet_cidrs" {
+  type        = list(string)
+  description = "Public Subnet CIDR values"
+  default     = ["172.31.1.0/24"]
```
nazar-pc, line 48: This isn't public though, what does this mean?

aws/devnet/variables.tf
```diff
@@ -0,0 +1,122 @@
...
+
+variable "private_subnet_cidrs" {
+  type        = list(string)
+  description = "Private Subnet CIDR values"
+  default     = ["172.31.2.0/24"]
+}
+
+variable "disk_volume_size" {
+  type    = number
+  default = 50
```
nazar-pc: I'd probably prefer to not have default value, but rather require this being specified for every node explicitly. It'll likely be different for different types of nodes anyway.

aws/network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -0,0 +1,153 @@
+locals {
+  bootstrap_nodes_ip_v4 = flatten([
+    #   [var.bootstrap-node-config.additional-node-ips],
+    [aws_instance.bootstrap_node.*.public_ip]
```
nazar-pc: Something left from older attempts

aws/network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -0,0 +1,153 @@
...
+  provisioner "file" {
+    source      = "./bootstrap_node_keys.txt"
+    destination = "/subspace/node_keys.txt"
+  }
+
+  # copy DSN bootstrap node keys file
+  provisioner "file" {
+    source      = "./dsn_bootstrap_node_keys.txt"
+    destination = "/subspace/dsn_bootstrap_node_keys.txt"
+  }
```
nazar-pc: I find it strange to copy all the keys to all nodes, which are not even removed afterwards. Even when node is pruned, the keys are still present and might get mixed with something later.

aws/network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -0,0 +1,153 @@
...
+      "echo DSN_LISTEN_PORT=${var.bootstrap-node-config.dsn-listen-port} >> /subspace/.env",
+      "echo NODE_DSN_PORT=${var.bootstrap-node-config.node-dsn-port} >> /subspace/.env",
+      "echo GENESIS_HASH=${var.bootstrap-node-config.genesis-hash} >> /subspace/.env",
+
+      # create docker compose file
+      "sudo chmod +x /subspace/create_compose_file.sh",
+      "sudo /subspace/create_compose_file.sh ${var.bootstrap-node-config.reserved-only} ${length(local.bootstrap_nodes_ip_v4)} ${count.index}",
+
+      # start subspace node
+      "sudo chown ubuntu:ubuntu /subspace/docker-compose.yml",
```
nazar-pc: Isn't it already owned by the user that runs it?

aws/network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -0,0 +1,153 @@
...
+      "sudo docker compose -f /subspace/docker-compose.yml up -d --build ",
```
nazar-pc: What are we building here? Generated `docker-compose.yml` doesn't seem to contain anything that needs to be built.

This seems to be the same kind of thing as installing various unused packages. Your code is also your documentation. When I see `--build`, I immediately start wondering why do we build anything just to discover that we in fact do not build anything.

aws/network-primitives/instances.tf
```diff
@@ -0,0 +1,297 @@
...
+
+  lifecycle {
+
+    create_before_destroy = true
+
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "sleep 180",
```
nazar-pc: :thinking: why? This is the kind of thing that typically randomly fails in practice.

aws/network-primitives/instances.tf
```diff
@@ -0,0 +1,297 @@
...
+  availability_zone = element(var.azs, count.index)
+  # Security Group
+  vpc_security_group_ids = ["${aws_security_group.gemini_sg.id}"]
+  # the Public SSH key
+  key_name                    = var.aws_key_name
+  associate_public_ip_address = true
+  ebs_optimized = true
+  ebs_block_device {
+    device_name = "/dev/sda1"
+    volume_size = 100
```
nazar-pc: These disk sizes look severely underestimated. Are you sure they are sufficient?

aws/network-primitives/network.tf
```diff
@@ -0,0 +1,193 @@
+resource "aws_vpc" "gemini_vpc" {
```
nazar-pc: Should it be called `gemini_*` if it is in primitives (here and below)?

aws/network-primitives/provider.tf
```diff
@@ -0,0 +1,32 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">=4.55.0"
+    }
+
+    cloudflare = {
+      source  = "cloudflare/cloudflare"
+      version = "~> 3.18.0"
```
nazar-pc: What is the reason to not upgrade minor version of CloudFlare dependency?

aws/network-primitives/network.tf
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+resource "aws_route_table_association" "private_route_table_subnets_association" {
+  count          = length(var.private_subnet_cidrs)
+  subnet_id      = element(aws_subnet.private_subnets.*.id, count.index)
+  route_table_id = element(aws_route_table.private_route_table.*.id, count.index)
+}
+
+resource "aws_security_group" "gemini_sg" {
```
nazar-pc: Looks like port 50000 is missing in this list despite used in DSN bootstrap node container

aws/network-primitives/scripts/install_docker.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+# updates
+export DEBIAN_FRONTEND=noninteractive
+sudo apt update -y
+sudo apt install -y curl jq
```
nazar-pc: `jq` doesn't seem to be used anywhere

aws/network-primitives/bootstrap_node_provisioner.tf
nazar-pc: Also why do we start the container with `docker-compose` if that should be done by systemd instead?

nazar-pc (commented):
Commit structure is a bit of a mess, I'd like to request you to re-read https://www.notion.so/subspacelabs/Pull-Request-Guide-cd8d0f0f5fe643d5bfde6145450d5ff6 (I have recently updated it, you likely didn't see the latest revision). It gives numerous suggestions how to make reviewers life easier that could be applied to this PR.

I left mostly questions, there was a lot of code duplication, so I might have missed something, but overall makes sense to me.

As to discussion about provisioning method and scripts.

I do agree that there was no need to copy-paste scripts. You can add `sudo` and it will continue working the same way on DO and Hetzner just fine. As to the user it also doesn't make a meaningful difference and container can be stared under `root`, permissions are lower in the container anyway.

So you count have unified them and that is possible to do in a separate PR such that diff here is smaller. I don't see a big reason to change them in a significant way even though we could simplify things for AWS/DO, systemd is only needed for Hetzner. The deployment method will change in the future, but there is no need to do all changes at once.

That said, please don't change to shared scripts in this PR, let's leave it as is, it'll require even more time to review thousands of lines of code. Just take this as a feedback for future changes.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: I'd prefer to not use systemd with terraform since it can break the run and would rather use systemd with configuration management. The --build flag just forces a rebuild of the images, not really applicable here since we don't use the `latest` tag, but it just forces a rebuild of the images before running them.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: It's just a precaution to ensure the ownership is set.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: This is remnants from the older version, I will modify the prune script to remove keys, but currently it would just copy over the existing ones.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: No, i just temporarily commented it, but it will be removed since we are not using another provider with this and using this strictly for AWS.

aws/devnet/variables.tf
DaMandal0rian: sure, will do that, I set these as a precaution of smaller values so mistakenly don't create large volumes and incur extra charges, the size of disks are set in the module configs and will overwrite this default.

aws/devnet/variables.tf
DaMandal0rian: this is just the IP address range to assign the nic cards for machines in the subnet, public interfaces are exposed on another interface or through the internet gateway or elastic IPs.

aws/devnet/variables.tf
DaMandal0rian: this is per zone.

aws/devnet/variables.tf
DaMandal0rian: These will be removed since we are strictly using AWS only, and have no need for these.

aws/devnet/main.tf
DaMandal0rian: I use it to distinguish between config variables and the global variables. However, our older networks have it more inconsistent. i.e

```
bootstrap-node-config = {
    droplet_size        = var.droplet-size
    deployment-version  = 
    regions             = []
    nodes-per-region    = 0
}
```

aws/network-primitives/instances.tf
DaMandal0rian: actually, I will set these in configs.

aws/network-primitives/instances.tf
DaMandal0rian: It gives the cloud-init scripts time to finish before trying to run the remote provisioners else it will timeout, the aws launches go though an initialization stage before the health checks are done and they system state is "ready/running".

aws/network-primitives/network.tf
DaMandal0rian: That is the name of the resource, the naming is correct and can't be set dynamically afaik.

aws/network-primitives/network.tf
DaMandal0rian: Will add this, actually now i think of it that port was never set in firewall from our DO declarations.

aws/network-primitives/provider.tf
DaMandal0rian: the `~>` will pull in the latest versions greater than 3.18.0. I have upgraded the provider version.

DaMandal0rian:
> Commit structure is a bit of a mess, I'd like to request you to re-read https://www.notion.so/subspacelabs/Pull-Request-Guide-cd8d0f0f5fe643d5bfde6145450d5ff6 (I have recently updated it, you likely didn't see the latest revision). It gives numerous suggestions how to make reviewers life easier that could be applied to this PR.
> 
> I left mostly questions, there was a lot of code duplication, so I might have missed something, but overall makes sense to me.
> 
> As to discussion about provisioning method and scripts.
> 
> I do agree that there was no need to copy-paste scripts. You can add `sudo` and it will continue working the same way on DO and Hetzner just fine. As to the user it also doesn't make a meaningful difference and container can be stared under `root`, permissions are lower in the container anyway.
> 
> So you count have unified them and that is possible to do in a separate PR such that diff here is smaller. I don't see a big reason to change them in a significant way even though we could simplify things for AWS/DO, systemd is only needed for Hetzner. The deployment method will change in the future, but there is no need to do all changes at once.
> 
> That said, please don't change to shared scripts in this PR, let's leave it as is, it'll require even more time to review thousands of lines of code. Just take this as a feedback for future changes.

I understand, I put the scripts into the new network primitives folder under the AWS folder, most of them copied over, and only 2 were actually modified. I didn't modify the others, the diff is just them being added to the path.

Additionally, i didn't want to spend too much time on these scripts since they will be removed in the future completely.

DaMandal0rian:
> I understand, I put the scripts into the new network primitives folder under the AWS folder, most of them copied over, and only 2 were actually modified. I didn't modify the others, the diff is just them being added to the path.

I understand, I put the scripts into the new network primitives folder under the AWS folder, most of them copied over, and only 2 were actually modified. I didn't modify the others, the diff is just them being added to the path.

> > Lastly, your last reply is very contradictive, do you you want to use the same scripts, change scripts to not assume root, use some other deployment tool?
> 
> Well, I would like for us to use the same scripts if we are taking that route. But you and @nazar-pc did not like that approach. If we going to use the ansible or other then we should not use these scripts and the deployment is done using the chosen tool completely. These scripts generate docker compose files and the chose tool should do it instead of these scripts.
> 
> But considering the changes you have done, seems like we are going to use the same approach we currently have, then the scripts needs to be updated to work AWS instead of creating a new set of scripts and rewrite is not necesary. Systemd can run without root as well.
> 
> > Not sure if a thesis is required to discuss why a deployment works with root user and wouldn't with ubuntu or any other user with existing scripts.
> 
> Again sarcasm doesn't help anyone. I'm merely suggesting an approach since you asked my review. I'm happy to sit out on this and let you and @nazar-pc handle this completely :)

I understand, I put the scripts into the new network primitives folder under the AWS folder, most of them copied over, and only 2 were actually modified. I didn't modify the others, the diff is just them being added to the path. These were just migrated over for now so we can launch faster, and will be removed completely in the future. If we are trying to seek perfection in every detail beforehand, rather than a work-in-progress with iterations, then it's going to just take longer to deliver things imo. Infra always changes and has improvements along the way.

aws/devnet/variables.tf
nazar-pc: Why is it called `public_subnet_cidrs` then if it is not public? I'm confused.

aws/devnet/variables.tf
nazar-pc: If they are set explicitly anyway, why having a default?

aws/network-primitives/bootstrap_node_provisioner.tf
nazar-pc: Hm... it makes little sense to me still. You've created it under current user, then using `sudo` to set ownership to current user :man_shrugging: 

It is either necessary or not, I don't like arbitrary "just in case" code that is not actually required.

aws/network-primitives/bootstrap_node_provisioner.tf
nazar-pc: Can we add a TODO then? It is easy to forget and others don't have that context in the first place.

aws/network-primitives/bootstrap_node_provisioner.tf
nazar-pc: If there is nothing to rebuild, it should not specify `--build`.

As to using systemd or not, since we use systemd right now, this line is incorrect since it will start container without systemd and will not apply iptables rules as the result.

aws/network-primitives/instances.tf
nazar-pc: Is there a way to detect when cloud-init finishes rather than having arbitrary timeouts?

180 seconds is A LOT, would be much better to proactively detect this rather than waiting for whole 3 minutes unnecessarily.

aws/network-primitives/instances.tf
nazar-pc: Well, you did, but as far as I see you have used those 50G sized that I complained about in the other comment. So we have kind of made a full circle now. Or am I missing something and those values come from somewhere else?

aws/network-primitives/network.tf
nazar-pc: I'd argue the name is not correct. This file is `aws/network-primitives/network.tf`, meaning it is used by all networks as a primitive, not just by Gemini networks, so should be named accordingly, otherwise looks misleading to me.

nazar-pc (commented):
> I understand, I put the scripts into the new network primitives folder under the AWS folder, most of them copied over, and only 2 were actually modified. I didn't modify the others, the diff is just them being added to the path.
>
> Additionally, i didn't want to spend too much time on these scripts since they will be removed in the future completely.

Understood, in this case a good practice is to copy them with 0 modifications in a separate commit (such that it is clear to reviewer that it doesn't need to be reviewed) and then doing modifications in a distinct commit, such that reviewer will spend little time if they are already familiar with the previous version. Copying and modifying at the same time means the whole file needs to be re-reviewed from scratch.

> These were just migrated over for now so we can launch faster, and will be removed completely in the future. If we are trying to seek perfection in every detail beforehand, rather than a work-in-progress with iterations, then it's going to just take longer to deliver things imo. Infra always changes and has improvements along the way.

Completely agree with you, this is why we need a separate commit that literally says something like "Files are copied from X with zero modifications, will be changes in the next commit" and maybe even add an explanation why did you decide to copy the files rather than reusing originals. What you explained makes sense to me, but it was not clear from the original PR, commits and description provided.

aws/devnet/variables.tf
DaMandal0rian: because that is the subnet that has public access to internet. The only thing that distinguishes pulic and private is that one has full ingress and egress and the private uses the public subnet nat gateway and internet gateway.  Please see https://docs.aws.amazon.com/vpc/latest/userguide/configure-subnets.html#subnet-basics

aws/devnet/variables.tf
DaMandal0rian: I have removed the default. The variable will be set in tfvars file to desired amount needed.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: Already have, and have added to the prune script already to remove key files. This was done in the prune_docker_system.sh script file.

aws/network-primitives/instances.tf
DaMandal0rian: There is no way for cloud-init to tell terraform it's finished. However, I have removed this and just increased the timeout for ssh connections.

aws/network-primitives/network.tf
DaMandal0rian: I just used "network" instead, and the "name" tag will set the actual network name to identify it.

aws/devnet/variables.tf
DaMandal0rian: When I remove it, we get the following:

│ Error: Missing required argument
│ 
│   on main.tf line 1, in module "devnet":
│    1: module "devnet" {
│ 
│ The argument "disk_volume_size" is required, but no definition was found.

Again, if the user needs to adjust this they can set the desired amount in the module itself if they need something higher than the default, or I'm happy setting a higher default.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: It is 100% necessary. Otherwise it's owned by root user.

`-rw-r--r--  1 root   root   2050 Jun  9 07:10 docker-compose.yml`

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: The Iptables and systemd will not be configured in terraform, this is not the way to go. Also, will those Iptable rules have effect on the aws instances?

`docker0: flags=4099<UP,BROADCAST,MULTICAST>  mtu 1500
        inet 172.17.0.1  netmask 255.255.0.0  broadcast 172.17.255.255

ens5: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 9001
        inet 172.31.1.139  netmask 255.255.255.0  broadcast 172.31.1.255`
        
   Secondly, those rules might be for older docker versions, but as of lately they recommend not to manually manipulate the rules.
   https://docs.docker.com/network/packet-filtering-firewalls/#add-iptables-policies-before-dockers-rules

Thirdly, the current rules do nothing since they are targeting the wrong chain and should be targeting `DOCKER-USER`

aws/network-primitives/instances.tf
DaMandal0rian: Actually,  the comment is outdated, not sure why you're still seeing that.

aws/devnet/variables.tf
nazar-pc: Well, this is exactly the problem I am trying to explain to you. Since this is a default value, it should not be required. If something else doesn't work after you remove default value, it seems to indicate you were in fact using default value, which as we already discussed is way too low and should be multiple times higher and specified explicitly.

aws/network-primitives/bootstrap_node_provisioner.tf
nazar-pc: Right, maybe remove `sudo` from above two commands instead? It seems to be happening backwards because of that.

aws/network-primitives/bootstrap_node_provisioner.tf
nazar-pc: I thought we are deploying the same thing we had with minimal changes to AWS. What you're describing is quite far from it. Moreover, you're saying it is not the way to go, yet this PR still deploys the same systemd service and iptables rules, but in contrast to deployment to Hetzner where everything was done properly and in correct order, the way you wrote it is not equivalent, hence this comment and request to correct it to the way it was before.

The rules will certainly have an effect. They are probably redundant, but I have not visibility into what you have actually changed here comparing to what we had before unless I check out code locally and do file comparison manually.

> Thirdly, the current rules do nothing since they are targeting the wrong chain and should be targeting DOCKER-USER

They were certainly working properly and were not related to Docker as such at all. Our servers were blocked on Hetzner because we were making requests to private IP ranges via public interfaces, that is what those rules were blocking and they certainly did the job.

aws/network-primitives/instances.tf
nazar-pc: This comment is outdated, but the issue is not, which is why I'm continuing this conversation. I see several places and none of them have the correct disk size. So to the best of my understanding we still have an issue even though code has changed since I wrote the first comment here.

aws/network-primitives/instances.tf
nazar-pc: > There is no way for cloud-init to tell terraform it's finished. However, I have removed this and just increased the timeout for ssh connections.

It doesn't need to. `cloud-init` is doing something, right?  So you can have a loop here that will check for whatever you expect cloud-init to finish doing every few seconds and move forward only when it is done.

Removing delay and increasing timeout makes even less sense now. Will it automatically retry or something?

nazar-pc (commented):
Please don't resolve every single comment just because you have replied to it. I have to expand every single one, reply to it and unresolve it manually after that. Only resolve those you're 100% were resolved and no questions there will remain. I will resolve those that I have no question about anymore myself.

aws/network-primitives/bootstrap_node_provisioner.tf
DaMandal0rian: It fails in that case

nazar-pc: The natural response to that would be how does it fail exactly?

What I see is that it reads some environment variables and writes text to a file. None of those things inherently require root access. I guess the issue might be if `/subspace` is created and owned by root user, but then why would you want the file in it to be owned by `ubuntu` user I do not understand.

DaMandal0rian: `provisioner "file"` uses scp to copy the files. It fails with permission denied for the user `ubuntu`  since the path  `/subspace` is directly under the `/` path.

https://github.com/hashicorp/terraform/issues/8238 

Alternatively, I can copy to a /home/ubuntu/subspace because the user `ubuntu` will have the right uid/gid set.

aws/network-primitives/domain_node_provisioner.tf
```diff
@@ -0,0 +1,197 @@
...
+    agent       = true
+    private_key = file("${var.private_key_path}")
+    timeout     = "300s"
+  }
+
+  # create subspace dir
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p /subspace",
+      "sudo mkdir -p /home/ubuntu/bin/",
```
DaMandal0rian: remove this

aws/network-primitives/farmer_node_provisioner.tf
```diff
@@ -0,0 +1,157 @@
...
+    agent       = true
+    private_key = file("${var.private_key_path}")
+    timeout     = "300s"
+  }
+
+  # create subspace dir
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p /subspace",
+      "sudo mkdir -p /home/ubuntu/bin/",
```
DaMandal0rian: remove

aws/network-primitives/rpc_node_provisioner.tf
```diff
@@ -0,0 +1,193 @@
...
+    agent       = true
+    private_key = file("${var.private_key_path}")
+    timeout     = "300s"
+  }
+
+  # create subspace dir
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p /subspace",
+      "sudo mkdir -p /home/ubuntu/bin",
```
DaMandal0rian: remove

aws/network-primitives/full_node_provisioner.tf
```diff
@@ -0,0 +1,154 @@
...
+    agent       = true
+    private_key = file("${var.private_key_path}")
+    timeout     = "300s"
+  }
+
+  # create subspace dir
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p /subspace",
+      "sudo mkdir -p /home/ubuntu/bin/",
```
DaMandal0rian: remove

aws/network-primitives/bootstrap_node_provisioner.tf
```diff
@@ -0,0 +1,152 @@
...
+    agent       = true
+    private_key = file("${var.private_key_path}")
+    timeout     = "300s"
+  }
+
+  # create subspace dir
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mkdir -p /subspace",
+      "sudo mkdir -p /home/ubuntu/bin/",
```
DaMandal0rian: remove

nazar-pc (approved):
Latest version makes much more sense, thanks!

PR #153, opened 2023-07-25
add subspace.net to terraform

DaMandal0rian:
- Add subspace.net  zone DNS records to terraform

### DIFF
--- a/resources/dns/mailserver_records.tf
+++ b/resources/dns/mailserver_records.tf
@@ -1,4 +1,4 @@
-// mailserver records
+// mailserver records for subspace.netwwork
 resource "cloudflare_record" "mailserver_mx_1" {
   name     = "subspace.network"
   comment  = "MX record pointing to our preferred mailserver"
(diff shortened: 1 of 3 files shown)


resources/dns/mailserver_records.tf
```diff
@@ -1,4 +1,4 @@
-// mailserver records
+// mailserver records for subspace.netwwork
```
vedhavyas:
```suggestion
// mailserver records for subspace.network
```

vedhavyas (commented):
LGTM!

need to fix the typo

resources/dns/mailserver_records.tf
```diff
@@ -111,4 +111,100 @@ resource "cloudflare_record" "mail" {
...
+  comment = "SPF record"
+  proxied = false
+  ttl     = 3600
+  type    = "TXT"
+  value   = "v=spf1 include:_spf.google.com -all"
+  zone_id = data.cloudflare_zone.cloudflare_zone_subspace_net.id
+}
+
+resource "cloudflare_record" "mail_net" {
+  name    = "mail"
```
nazar-pc: This one is not in CloudFlare and I don't think we need it. `subspace.net` is just an alias, it is not a standalone domain name.

If you still think it should be added then PR description should contain explanation for this. So far I do not understand why, the goal was to simply add what we have in CloudFlare to Terraform.

DaMandal0rian: You are right, actually I was thinking why we even need it for the other domain too `subspace.network`,  I included it because i thought it may be needed, but i have removed it

nazar-pc (approved):

PR #161, opened 2023-09-05
Cloudflare loadbalancers

### LINKED ISSUE #162: Load-Balanced RPCs

DaMandal0rian:
This PR creates cloudflare loadbalancers to distribute traffic between RPC and EVM domain origins.
It contains:
- health monitor checks (currently TCP, will add HTTP(S) when we expose a metrics endpoint)
- loadbalancer pools
- loadbalancers

closes #162


cloudflare-lb/variables.tf
```diff
@@ -0,0 +1,27 @@
...
+  sensitive   = true
+}
+
+variable "domain" {
+  default = "subspace.network"
+}
+
+variable "network" {
+  type        = string
+  description = "gemini network name"
```
vedhavyas: Shouldn't this be a variable so that we can re-use for multiple networks ?

DaMandal0rian: @vedhavyas not sure I follow, this is a variable. If you mean a data-type to support multiple networks, I think we only need to focus on the current live network. This will later be refactored to a module.

cloudflare-lb/main.tf
```diff
@@ -0,0 +1,155 @@
...
+  zone_id  = data.cloudflare_zones.subspace.id
+  hostname = "rpc.${var.network}.${var.domain}"
+  type     = "A"
+}
+
+resource "cloudflare_record" "rpc-0" {
+  count   = length(data.cloudflare_record.existing_rpc_record.hostname) > 0 ? 0 : 1
+  zone_id = var.zone_id
+  name    = "rpc"
+  value   = "52.91.27.239"
```
vedhavyas: are there the load balancer IPs cloudfalre provided or the actual rpc IPs ?
If it is the latter, then shouldn't they come from the terrform config instead ?

cloudflare-lb/variables.tf
vedhavyas: No, i think this should already be a module that we can use for multiple networks.

DaMandal0rian:
@vedhavyas this has changed into a module now. To answer your question:

> are there the load balancer IPs cloudfalre provided or the actual rpc IPs ?
>If it is the latter, then shouldn't they come from the terrform config instead ?

The IPs are the actual RPCs. Since using a module I have set it dynamically.

cloudflare-lb/cloudflare_lb_module/variables.tf
```diff
@@ -0,0 +1,133 @@
...
+    name     = string
+    hostname = string
+    value    = string
+    type     = string
+    tags     = list(string)
+  }))
+  default = [
+    {
+      name     = "rpc-0"
+      hostname = "rpc.gemini-3f.subspace.network"
```
vedhavyas: Shouldn't these be coming from the project that uses this module instead ?
Seems weird seeing gemini-3f as default under a terraform module

vedhavyas (commented):
Make sense but terraform module should not default to any specific network but rather those values should come from the project that uses this module

cloudflare-lb/cloudflare_lb_module/variables.tf
DaMandal0rian: Yeah this has been removed, missed it when refactoring to module.

vedhavyas (approved):

Issue #142, opened 2023-06-08
Make terraform code DRY and create modules


DaMandal0rian:
The task is to make our terraform concise and reusable, use modules to make it DRY and composable.

DaMandal0rian:
1. I have removed the duplicated base module for explorers and made one under templates now.
2. telemetry has been refactored to use the aws module in templates now and removed the base module
3. network primitives uses the aws module now for AWS instances but keeps most functionality intact
4. Templates folders now hold the root modules, and going forward resources will use these modules to keep the declarations DRY.
5. Created a DNS records module, the current dns.tf files will in various resources could be refactored later to use this, but not urgent priority right now. Going forward they will be used and older resources will be destroyed anyway.

DaMandal0rian:
Still need to do the hetzner module and also refactor AWS testing framework once #263 and #271 are merged. Will do the hetzner module in #255

PR #175, opened 2023-09-25
Ephemeral testing framework

### LINKED ISSUE #167: ephemeral devnets with AWS

DaMandal0rian:
This PR creates the ephemeral testing framework to spin up "devnets" for testing purposes.
- aws declarations
- hetzner declarations
- make docker compose files build from branch

Instances can spin up on AWS or re-use hetzner resources.

closes #167

### DIFF
--- a/testing-framework/ec2/base/ami.tf
+++ b/testing-framework/ec2/base/ami.tf
@@ -1,0 +1,20 @@
+data "aws_ami" "ubuntu_amd64" {
+  most_recent = true
+
+  filter {
+    name   = "name"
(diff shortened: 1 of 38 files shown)


vedhavyas (commented):
We seems to not re-use terraform modules. Can we please do that? Ideally, once module should work for both Hetzner and AWS. If not possible two different versions of this though will look ugly but definately not copying the base for every project IMO

DaMandal0rian:
@vedhavyas we can't re-use AWS and hetzner together, they use different users and key formats etc. There will be a different task to re-write most terraform to use modules in #142

vedhavyas (approved):
Assuming you have copied from the exisiting network primitives

PR #183, opened 2023-10-03
Add CI/CD workflows for terraform automation

### LINKED ISSUE #184: Github CI pipelines for automation of networks with terraform
```[tasklist]
### Tasks
- [x] deploy gemini-3f with CI and terraform automation
- [x] deploy devnet with CI and terraform automation
- [x] deploy ephemeral devnets with CI and terraform automation
- [ ] https://github.com/subspace/infra/pull/183
```

DaMandal0rian:
The PR adds terraform automation with github actions CI for devnet, gemini-3(x) and ephemeral devnets.

Closes #184


DaMandal0rian:
The checks will fail since other PRs need to be merged beforehand.

.github/workflows/devnet_main_deploy.yml
```diff
@@ -0,0 +1,27 @@
+name: Main CD for Devnet Deployment
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+  push:
+    paths:
+      - "aws/devnet/**"
```
nazar-pc: So it will redeploy regardless of things changing if merged to `main`, which is probably fine, but also why?

But it will also redeploy on push to any branch whatsoever that changes `aws/devnet/**`, which I don't think is what we want.

Similar comment applies to all other workflows in this PR.

.github/workflows/devnet_main_deploy.yml
```diff
@@ -0,0 +1,27 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    uses: ./.github/workflows/terraform_template_deploy.yml
```
nazar-pc, line 20: Why `./.github` and not just `.github`?

.github/workflows/terraform_gh_runner.yml
```diff
@@ -1,16 +1,18 @@
...
 on:
-  push:
+  workflow_dispatch:
+  pull_request:
     branches:
       - main
     paths:
-      - './github-runners/terraform/base/**'
-  workflow_dispatch:
+      - "./github-runners/terraform/base/**"
```
nazar-pc: Similarly here `./` seems unnecessary

.github/workflows/terraform_gh_runner.yml
```diff
@@ -1,16 +1,18 @@
...
 
 jobs:
   terraform_gh_runner:
     runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
nazar-pc, line 15: It doesn't seem to be used, why setting?

.github/workflows/terraform_gh_runner.yml
```diff
@@ -19,40 +21,47 @@ jobs:
...
+          terraform_version: "1.5.7"
           cli_config_credentials_token: ${{ secrets.TF_CLOUD_TOKEN }}
 
       - name: Install dependencies
         run: |
           # Install any dependencies required by your Terraform code
 
       - name: Run Bash Script
+        env:
+          GH_TOKEN: ${{ github.token }}
```
nazar-pc: You're using `secrets.GITHUB_TOKEN` above and `github.token` here, which are both the same thing. Would be nice to make it consistent.

But this environment doesn't seem to be used either.

.github/workflows/terraform_gh_runner.yml
```diff
@@ -19,40 +21,47 @@ jobs:
...
         run: |
-            # GitHub repository and access token
-            repo="subspace/infra"
-            token=${{ secrets.PAT_TOKEN }}
+          # GitHub repository and access token
+          repo="subspace/infra"
+          token=${{ secrets.PAT_TOKEN }}
```
nazar-pc, line 39: A third token, can you explain what this third token is and why do we need it? I would have expected that workflow should be able to use `secrets.GITHUB_TOKEN`/`github.token` instead.

.github/workflows/terraform_gh_runner.yml
```diff
@@ -19,40 +21,47 @@ jobs:
...
 
-            # Set the runner token as an environment variable
-            export RUNNER_TOKEN="$runner_token"
+          # Set the runner token as an environment variable
+          export RUNNER_TOKEN="$runner_token"
 
-            # Store the token as a secret in GitHub Actions
-            gh secret set RUNNER_TOKEN -r "$repo" -b "$runner_token"
+          # Store the token as a secret in GitHub Actions
+          gh secret set RUNNER_TOKEN -r "$repo" -b "$runner_token"
```
nazar-pc: This makes little sense, why are we setting it as environment variable in two different ways AND secrets? We only seem to use the first environment variable in the next step.

All this token management looks like a complete mess to me. A bunch of random things, some of which are used and some are not, but not clear why.

Comments were supposed to help, but these are exactly the kinds of comments that shouldn't be written in the first place: they say exactly the same thing that code does, so I essentially read the same information twice, which isn't helpful. Comments should add useful information, explaining why things are done, how they are useful but none of this is currently present.

.github/workflows/devnet_main_deploy.yml
DaMandal0rian: It will only run a plan, the apply/destroy will only run if input is set to yes.
https://github.com/subspace/infra/pull/183/files#diff-31842ccef98d0aadb01a72552ac62f9015e5139c67ebc0a9bc7d65ef272a1850R25-R32

So lets say i create a new branch and push to remote, it will show me the plan.

.github/workflows/devnet_main_deploy.yml
DaMandal0rian: makes sense will fix

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: The env variable needs to be set when calling github api, otherwise an error is thrown.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: It is used with github api, will make consistent. It is needed for api calls to get the PAT token for github runner registration.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: PAT token to interact with repo for creating a token for github runner registration. The `secrets.GITHUB_TOKEN` token can't be used for that purpose.

.github/workflows/devnet_main_deploy.yml
DaMandal0rian: It will complain in IDE that it can't find the re-usable workflow

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: There are essentially 3 tokens, `secrets.GITHUB_TOKEN` for using github api. `PAT_TOKEN` to access repository and create a runner registration token `runner_token` to register a runner for github actions. Yeah removed the extra step of setting it again since stored in  $GITHUB_ENV, was running as a bash script before and forgot to remove that part.

.github/workflows/terraform_gh_runner.yml
nazar-pc: What and where calls those APIs? I don't see any usage of it.

.github/workflows/terraform_gh_runner.yml
nazar-pc: GitHub API again? Which API, I don't see any usage of it.

.github/workflows/terraform_gh_runner.yml
nazar-pc: So you're using personal token for that? Is that a recommended approach? Seems fragile and generally problematic to me.

.github/workflows/terraform_gh_runner.yml
nazar-pc: But why do you store it in the environment variable at all? It is already stored in secrets. Or why is it stored in secrets if it is only necessary in environment variable for the next step. Also storing it in enviromnent variable probably means it will be printed in CI run in logs, which is a security issue.

This is why it looks like a mess.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: The GITHUB_TOKEN is needed in CI or bash cli for API calls. The API is called with curl via https://github.com/subspace/infra/pull/183/files/99a5ac25712fdda531fd88026a12f96aafeacb83#diff-b46b68b6df852ad5f8fc96162f55c2fe198f6d53eb2fe7d7ce12fcd4b2650ba6R39-R43

> If you want to use the API in a GitHub Actions workflow, GitHub recommends that you authenticate with the built-in GITHUB_TOKEN

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: https://github.com/subspace/infra/pull/183/files/99a5ac25712fdda531fd88026a12f96aafeacb83#diff-b46b68b6df852ad5f8fc96162f55c2fe198f6d53eb2fe7d7ce12fcd4b2650ba6R39-R43

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: Yes, it's the recommended approach https://docs.github.com/en/rest/guides/getting-started-with-the-rest-api?apiVersion=2022-11-28&tool=curl#authenticating

GH_TOKEN does not have enough permissions on the repo/org level.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: Storing as the env variable in that step will not print it to CI, its just echoing the string into GITHUB_ENV. However I have removed it and just used the one from the secret at https://github.com/subspace/infra/pull/183/files#diff-b46b68b6df852ad5f8fc96162f55c2fe198f6d53eb2fe7d7ce12fcd4b2650ba6R63

.github/workflows/terraform_gh_runner.yml
nazar-pc: What you linked doesn't use `GITHUB_TOKEN` environment variable, it uses `$token` that is obtained in a different way.

.github/workflows/terraform_gh_runner.yml
nazar-pc: How is linked code related to `GH_TOKEN`?

.github/workflows/terraform_gh_runner.yml
nazar-pc: Why still doing `export RUNNER_TOKEN="$runner_token"`?

What about race conditions, it is possible that there are multiple tokens issued concurrently in different jobs. I'm also wondering if there is a limit for number of these tokens or whether there is an expiration limit for them. Because if there is a limit we need to reuse them, if there is no expiration we can probably just store them in secrets once and never re-request.

nazar-pc (commented):
CI failed with `Error: Invalid command-line option`, which I think means something needs to be fixed.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: That was removed earlier the export statement, shouldn't be there if you check the file. PAT tokens can set an expiry date. Tokens that have expired can be regenerated, giving you a duplicate token with the same properties as the original. It's good practice to rotate though.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: Ok, for $token in the auth header of the curl request that is the pat token to authenticate to the repo and get the runner token. If i remove GH_TOKEN you will see an error in ci that you need to have GH_TOKEN to call github apis from workflow :). I wasn't using it before, but added it to fix the problem. Here is an old version:
https://github.com/subspace/infra/blob/bd296590f29cdd730ab29869c7ade1258cb395ae/.github/workflows/terraform_gh_runner.yml

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: Ok, for $token in the auth header of the curl request that is the pat token to authenticate to the repo and get the runner token. If i remove GH_TOKEN you will see an error in ci that you need to have GH_TOKEN to call github apis from workflow :). I wasn't using it before, but added it to fix the problem. Here is an old version:
https://github.com/subspace/infra/blob/bd296590f29cdd730ab29869c7ade1258cb395ae/.github/workflows/terraform_gh_runner.yml

nazar-pc: Then show me where it is used :man_shrugging: You have linked to `curl` request that doesn't use this environment variable as a reason for this environment variable to exist. It makes no sense to me.

.github/workflows/terraform_gh_runner.yml
nazar-pc: I am asking about `$runner_token` that this workflow stores in secrets, not PAT.

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: @nazar-pc removed it, here you go:
https://github.com/subspace/infra/actions/runs/6484473796/job/17608384650#step:5:24

.github/workflows/terraform_gh_runner.yml
DaMandal0rian: The token expires after one hour.
https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#create-a-registration-token-for-an-organization

.github/workflows/terraform_gh_runner.yml
nazar-pc: Okay, so it is not for `curl`, it was for `gh` CLI command, now it makes sense.

With that knowledge we understand why it is necessary. I think what we should do now in a secure way without using heavy tools like setting repo (I'm fairly certain simply setting environment variable wasn't) without using heavy tools like setting repo's secrets from workflow is to use outputs and mask them such that they are not visible in logs: https://github.com/orgs/community/discussions/25225#discussioncomment-3246942

That is the goal here: to pass the token from one step into another. What we had in earlier versions of this PR are various suboptimal/incorrect ways of achieving that ultimate goal.

DaMandal0rian: correct, github cli, I should have been more clear about that "The GITHUB_TOKEN is needed in CI or bash cli for API calls" but couldn't remember exactly since it was a while ago. That being said the solution of masking the secret and just passing it is nice if using the ephemeral runners maybe but this solution and workflow is for the dedicated runners, where i need to retain the secret so I can unregister and delete runner if need be and remove it from github. See https://docs.github.com/en/free-pro-team@latest/rest/actions/self-hosted-runners?apiVersion=2022-11-28#delete-a-self-hosted-runner-from-an-organization

nazar-pc: But you mentioned that tokens only live for 1 hour, so you'll have to retrieve fresh token anyway. Why retaining it then at all?

DaMandal0rian: The token is not needed to deregister, a force delete can be done. So i've used the masking technique.

vedhavyas:
@DaMandal0rian is this ready to be reviewed? Looks like GH actions are still failing

DaMandal0rian:
> @DaMandal0rian is this ready to be reviewed? Looks like GH actions are still failing

@vedhavyas yes the PR is ready. The reason is the tfvars file does not exist in the path, but i will upload the encrypted version in another PR for the resources. I've also made some changes.

DaMandal0rian:
tfvars file is added in PR https://github.com/subspace/infra/pull/226/files

vedhavyas (commented):
I maybe be missing something here but looks like we can only run one ephemeral devnet at a time since TF uses the same working directory. This would also be a problem if multiple people want to deploy at the same time.

If its aws, can we deploy new instances instead ?
As for the hetzner, we should not allow new deployments unless the workspace is already destroyed so that machine can be re-used.

Though I'm not sure if this is out of scope for this PR.

DaMandal0rian:
@vedhavyas For hetzner can run the destroy input with workflow dispatch event. Also, please check the readme file which addresses this. This is outside the scope of this PR which deals with CI/CD. In another PR i can add a random generator to the backend (AWS) for `terraform` changes, but would rather dev's modify the backend and open a PR as instructions state in readme.

0xjeremyfrank:
@DaMandal0rian it has been awhile since this PR has progressed? Are there any updates?

vedhavyas:
Any updates on this one? Can we close it and open when it is ready ?

PR #185, opened 2023-10-03
DNS records changes

DaMandal0rian:
- add continuim.cc zone
- add subdomains for subspace.network
- renaming resources to follow convention

### DIFF
--- a/resources/dns/continuim-records.tf
+++ b/resources/dns/continuim-records.tf
@@ -1,0 +1,17 @@
+resource "cloudflare_record" "continuim_cc" {
+  name    = "continuim.cc"
+  proxied = false
+  ttl     = 3600
+  type    = "CNAME"
+  value   = "proxy-ssl.webflow.com"
+  zone_id = data.cloudflare_zone.continuim_cc.id
+}
+
+resource "cloudflare_record" "continuim_cc_www" {
+  name    = "www"
+  proxied = false
+  ttl     = 3600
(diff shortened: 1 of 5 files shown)


resources/dns/continuim-records.tf
```diff
@@ -0,0 +1,17 @@
+resource "cloudflare_record" "main" {
```
vedhavyas: maybe replace main with contunuim_cc ?

resources/dns/records.tf
```diff
@@ -75,7 +71,40 @@ resource "cloudflare_record" "uptime" {
   ttl     = 1
   type    = "A"
   value   = "35.89.37.220"
-  zone_id = data.cloudflare_zone.cloudflare_zone.id
+  zone_id = data.cloudflare_zone.subspace_network.id
+}
+
+// subspace Grafana monitor
+resource "cloudflare_record" "grafana" {
```
vedhavyas, line 78: If we have grafana exposed, do we have the auth setup on grafana ?

vedhavyas (commented):
Make sense. Do we have auth enabled for grafana and other services we are exposing here ?

resources/dns/records.tf
DaMandal0rian: https://grafana.subspace.network/

vedhavyas (approved):

PR #188, opened 2023-10-24
revert gateway and add playground endpoint

DaMandal0rian:
- fix cors for HEAD requests
- add playground endpoint
- revert api endpoint #187


dnoishi (approved):
LGTM

PR #189, opened 2023-10-24
add traefik proxy and replace caddy

DaMandal0rian:
remove caddy as a proxy and use traefik instead


ansible/network/files/docker-compose-rpc.yml
```diff
@@ -37,30 +40,52 @@ services:
...
+      - 443:443
     volumes:
-      - /var/run/docker.sock:/var/run/docker.sock:ro
-      - caddy_data:/data
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ./letsencrypt:/letsencrypt
 
   archival-node:
-    image: ghcr.io/${NODE_ORG}/node:${NODE_TAG}
+    image: ghcr.io/\${NODE_ORG}/node:\${NODE_TAG}
```
vedhavyas: I dont think this is correct, if you escape these envs, the image will be used as is and docker download fails

DaMandal0rian: @vedhavyas good catch thanks, missed removing when copying it. The other`docker-compose-domain.yml` is fine.

vedhavyas (approved):

PR #192, opened 2023-10-27
Gemini 3g

DaMandal0rian:
- add gemini-3f terraform infra
- update snapshot image
- add evm domain operator port
- enable UDP and use quic protocol
- change Domains DNS name to nova


aws/network-primitives/scripts/create_bootstrap_node_compose_file.sh
```diff
@@ -51,7 +51,8 @@ services:
     environment:
       - RUST_LOG=info
     ports:
-      - "30533:30533"
+      - "30533:30533/tcp"
+      - "30533:30533/udp"
```
nazar-pc: UDP should come first as higher priority as mentioned in Slack for consistency with priority

aws/network-primitives/scripts/create_bootstrap_node_compose_file.sh
```diff
@@ -63,7 +64,7 @@ services:
       - "--keypair"
       - \${DSN_NODE_KEY}
       - "--listen-on"
-      - /ip4/0.0.0.0/tcp/30533
+      - /ip4/0.0.0.0/udp/30533/quic-v1
```
nazar-pc, line 67: As mentioned in Slack, this should listen on both UDP and TCP, you just replaced TCP with UDP

nazar-pc (changes requested):

aws/network-primitives/scripts/create_bootstrap_node_compose_file.sh
DaMandal0rian: Oh right, didn't know can use both simultaneously with address parameters, thought it was for preference to udp port.

aws/gemini-3g/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  farmer-node-config = {
+    instance-type          = var.instance_type["farmer"]
+    deployment-version     = 0
+    regions                = var.aws_region
+    instance-count         = var.instance_count["farmer"]
+    docker-org             = "subspace"
+    docker-tag             = "gemini-3g-2023-oct-31"
+    reserved-only          = true
+    prune                  = false
+    plot-size              = "10G"
```
nazar-pc: Didn't we deply with 2G instead?

nazar-pc (dismissed):
Just one question, makes sense overall

aws/gemini-3g/main.tf
DaMandal0rian: Yes, but i had already deployed to servers before you asked to deploy with 2GB. I manually changed it on the server, do we want to use 2GB going forward?

nazar-pc: Actually yes, I think it should be sufficient to bootstrap the network. I just don't like when things deployed to servers and in `main` diverging.

PR #198, opened 2023-10-31
changes for astral

DaMandal0rian:
- add monitoring netdata and newrelic
- change endpoints to nova from evm
- add variables
- fix postgres config mounting for squid


dnoishi (approved):
LGTM

PR #199, opened 2023-11-01
Add operator flag to first domain node

DaMandal0rian:
add the `--operator` and `keystore-path` flags for domain and port configuration to devnet too.

### DIFF
--- a/aws/devnet/main.tf
+++ b/aws/devnet/main.tf
@@ -32,6 +32,7 @@ module "devnet" {
     genesis-hash       = ""
     dsn-listen-port    = 30533
     node-dsn-port      = 30433
+    operator-port      = 40333
     disk-volume-size   = var.disk_volume_size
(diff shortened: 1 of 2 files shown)


aws/devnet/main.tf
```diff
@@ -32,6 +32,7 @@ module "devnet" {
     genesis-hash       = ""
     dsn-listen-port    = 30533
     node-dsn-port      = 30433
+    operator-port      = 40333
```
vedhavyas, line 35: I don't see where this var is used in the scripts

DaMandal0rian: @vedhavyas https://github.com/subspace/infra/blob/main/aws/network-primitives/variables.tf#L187
and was introduced in #192

NingLin-P (approved):
LGTM

PR #202, opened 2023-11-07
add support for multiple domain operators

DaMandal0rian:
- re-enable multiple operators 2 to be exact
- append domain id to base path
- append node id to keystore directory to distinguish operator keystores


NingLin-P (approved):
Thanks!

PR #216, opened 2023-11-20
add continuum.co DNS records

DaMandal0rian:
create dns records for continuum.co DNS zone

### DIFF
--- a/resources/dns/continuum-records.tf
+++ b/resources/dns/continuum-records.tf
@@ -1,0 +1,17 @@
+resource "cloudflare_record" "continuum_co" {
+  name    = "continuum.co"
+  proxied = false
+  ttl     = 3600
+  type    = "CNAME"
+  value   = "apex-loadbalancer.netlify.com"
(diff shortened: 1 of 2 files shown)


resources/dns/continuum-records.tf
```diff
@@ -0,0 +1,17 @@
...
+  value   = "apex-loadbalancer.netlify.com"
+  zone_id = data.cloudflare_zone.continuum_co.id
+}
+
+resource "cloudflare_record" "continuum_co_www" {
+  name    = "www"
+  proxied = false
+  ttl     = 3600
+  type    = "CNAME"
+  value   = "continuum-collective.netlify.app"
```
nazar-pc, line 15: I wonder why this one is different

nazar-pc (approved):

resources/dns/continuum-records.tf
DaMandal0rian: One is for the apex domain and one is for subdomains, i guess they route each to different proxies.

PR #219, opened 2023-11-24
add readme files

DaMandal0rian:
add readme for network and ephemeral devnets deployment with terraform


nazar-pc (approved):

aws/README.md
```diff
@@ -0,0 +1,133 @@
...
+   ```SH
+   terraform apply current-plan.tfplan
+   ```
+
+Terraform will apply changes and generate/update the **.tfstate** file.
+Be aware that state files can contain sensitive information. Do not expose it to the public repository.
+
+## Installation via Github Actions
+
+(coming soon)
```
vedhavyas, line 133: Can we also add this documentation already?
This is kind of an immediate requirement IMHO since this would allow us to use Gh actions to deploy a branch

PR #231, opened 2024-01-03
bump gemini-3g release jan-03

DaMandal0rian:
upgrade gemini-3g to release 2024-jan-03


nazar-pc (approved):
Note that images are not built yet

PR #243, opened 2024-01-30
add IPv6 support to network

DaMandal0rian:
The PR adds IPv6 support to the AWS VPC, subnets, route tables and VM instances. The subspace network configuration is also configured to listen on IPv6 to improve support/connectivity to the network.

closes #239

PR #235, opened 2024-01-12
update parameters for consensus and domain nodes

DaMandal0rian:
This PR updates the parameters for consensus node and domain nodes respective of the cli changes in [#2408](https://github.com/subspace/subspace/pull/2408) and [#2412](https://github.com/subspace/subspace/pull/2412)

closes https://github.com/subspace/infra/pull/236

Major changes:
- adds a module `network-primitives-archive` to backport support older networks, i.e `gemini-3g` and `gemini-3f` since the upcoming CLI changes support next iteration of devnet and gemini-3h and has breaking changes to the existing module docker-compose deployments.
- Add domain names instead of IPs for domain RPC/bootstrap nodes(https://github.com/subspace/infra/pull/235/commits/20f570874052d0a7afcbbea89fda6d45a661b318)

PR #245, opened 2024-02-01
add gemini-3h module and refactor older gemini-3x modules

### LINKED ISSUE #246: Infra for gemini-3h
```[tasklist]
### Tasks
- [ ] https://github.com/subspace/infra/pull/245
- [ ] https://github.com/subspace/infra/pull/243
- [ ] https://github.com/subspace/infra/pull/235
- [ ] https://github.com/subspace/infra/pull/247
```

DaMandal0rian:
This PR has 2 major changes for infrastructure and some minor changes.
1- Add gemini-3h module with latest arguments and POT entropy seed argument. It also changes the networking in terms of UDP/Quic and TCP multiaddrees format as read from the structured flat files, and uses the correct L4 network protocols respective to substrate and subspace.

Substrate addresses are configured with TCP and subspace with Quic as first priority followed by TCP.

2- The older networks are incompatible with gemini-3h and latest devnet, so gemini-3g and gemini-3f will have archived network primitives modules each, with respective features and parameters. 

3- Fixes traefik proxying for RPCs and docker networking.

4- Change bootstrap nodes instance types

5- Add IPv6 DNS records for some nodes

The PR improves on #235 #243.

*NOTE*: All modules have been tested and confirmed are working. For convenience only the gemini-3h module and new network primitives needs reviewing 

closes #246


aws/network-primitives/dns.tf
```diff
@@ -20,13 +20,38 @@ resource "cloudflare_record" "rpc" {
...
+  count   = length(local.domain_nodes_ip_v4)
   zone_id = data.cloudflare_zone.cloudflare_zone.id
   name    = "${var.domain-node-config.domain-prefix}.${var.network_name}"
-  value   = local.domain_node_ip_v4[count.index]
+  value   = local.domain_nodes_ip_v4[count.index]
   type    = "A"
 }
 
+resource "cloudflare_record" "core-domain_ipv6" {
+  count   = length(local.domain_nodes_ip_v4)
```
DaMandal0rian, line 31: If you see the count using ipv4 value and ipv6 for the dns record value, this is not a mistake. AWS assigns and instance 2x IPv6 records for the interface, and if not done this way it will create extra AAAA records and bootstrap records. we will end up have `bootstrap-0..4` AAAA records otherwise.

vedhavyas (commented):
Apologies for a delayed feedback. 

I'm not sure I understand the actual reason behind creating yet another network-primitives. 

> The older networks are incompatible with gemini-3h and latest devnet, so gemini-3g and gemini-3f will have archived network primitives modules each, with respective features and parameters.

I'm concerned with this approach of creating a new slighly changed versions. Initial Idea of creating network-primitives is to modularize is to use as a single module for all the networks. I understand, as we make more progress, there are breaking changes to the way network is initialized. Keeping that in mind, we have the config exposed to each network on how its needs to be initialized and terraform supports such conditionals and so does the bash scripts. From a top level view, the process is still the same for all the network albeit cli commands. I do see bash script are getting too configurable and we should maybe re-write those bash scripts in some other language that makes the upcoming changes to network much more easier to update while keeping it backward compatible. We dont necessaarily need to keep all the backward compatibility for only upto last 2 network if not less since devnet is always following the most upto date network anyway.

I would prefer if we do not add another network-primitives module but just keep only one that is used across all networks with a rich configurability for future changes as well

DaMandal0rian:
@vedhavyas is this doable right now considering the time-frame? We have https://github.com/orgs/subspace/projects/18/views/1?pane=issue&itemId=30258135 for this purpose. We also have to launch gemini-3h nova too.

Also, using conditionals everywhere will not scale and become tedious to manage. Using own modules is cleaner imo. We also want to keep the infra for launching gemini-3f and 3g if necessary. I think from gemini-3h it can be more configurable going forward, but with IPv6 changes and the cli changes not worth making it backward compatible with the current older networks.

vedhavyas (approved):
I'll leave the decision to you and @nazar-pc  in this case. Adding new modules with similar changes are generally not preferred though

nazar-pc:
I think I'm fine copying stuff in some cases, with highly conditional code is it easy to break old networks and have dead code hanging around

vedhavyas:
> with highly conditional code is it easy to break old networks and have dead code hanging around

I dont anticipate the module to have too many breaking changes between each network and we dont necessarily need to keep the compaitbility code but maybe for lastest 2 networks.

Either way, I do not have strong stance on this topic atm. If it makes it easier to manage to then I guess it is fine

PR #248, opened 2024-02-02
update explorer docs

DaMandal0rian:
update the nginx docs for subspace archive and squild with fixes to api endpoint


dnoishi (approved):
LGTM

PR #259, opened 2024-03-01
separate modules to templates directory

DaMandal0rian:
This is part 1 of refactoring the terraform modules to be more composable.

- adds the reusable aws modules for (VPC and EC2) resources
- moves hetzner module
- moves network primitives

removal and restructuring of resources are done:

- Move scripts and configs folders into templates module to be reusable.
- move old `network-primitives-archive` into templates module folder.
- remove current network-primitives module and add it to the templates root

The AWS EC2/VPC modules are new and will be reusable by explorer, gemini networks and any other resources going forward.

Hetzner module will also be shared by some resources, but will be refactored with composable scripts for null resources and latest updates for CLI arguments and compatibility with gemini-3h and devnet.

Current `network-primitives` are moved into the templates directory and will be shared by the networks child modules.


nazar-pc (commented):
5818 lines of added code and nothing removed. This is not a refactoring, where all of this is coming from?

DaMandal0rian:
> 5818 lines of added code and nothing removed. This is not a refactoring, where all of this is coming from?

removal is done here https://github.com/subspace/infra/pull/260

The AWS EC2/VPC modules are new and will be reusable by explorer, gemini networks and any other resources.

nazar-pc:
This is extremely confusing and feels like the same issue we're hitting over and over again. Can't you convert one small thing at a time, such that it is clear what is changing, were and how? Dumping a bunch of code in one PR and removing a bunch of code in another PR is not review-friendly way to do development.

nazar-pc:
> removal is done here #260
> 
> The AWS EC2/VPC modules are new and will be reusable by explorer, gemini networks and any other resources.

Why is this information not in the original PR description then? you need to explain what you are doing, how, why and what are the next steps. See https://github.com/subspace/subspace/pull/2581 as one of the latest PRs I have created for an example of what it should look like.

DaMandal0rian:
Only `templates/terraform/aws` needs to be reviewed. The rest are basically copy/paste into the templates folder. Does this need multiple PRs? I tried breaking it down by adding in one and removing in another to reduce the amount of files changed. However, since it's apparently hard to follow that, i have merged it into one and the commits show what is happening. There are many further changes to go but those will be in smaller changes, this is the major one since it moves different modules to a new reusable modules folder.

PR #261, opened 2024-03-03
add reusable AWS module

DaMandal0rian:
This is part 1 of refactoring the terraform modules to be more composable. The PR adds a common templates directory and AWS modules.

- adds the reusable aws modules for (VPC and EC2) resources

closes #142

### DIFF
--- a/templates/terraform/aws/ec2/ami.tf
+++ b/templates/terraform/aws/ec2/ami.tf
@@ -1,0 +1,20 @@
+data "aws_ami" "ubuntu_amd64" {
+  most_recent = true
+
+  filter {
+    name   = "name"
+    values = ["ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-*"]
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "architecture"
+    values = ["x86_64"]
+  }
+
+  owners = ["099720109477"]
+}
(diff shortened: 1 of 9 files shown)


templates/terraform/aws/ec2/ami.tf
```diff
@@ -0,0 +1,20 @@
...
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "architecture"
+    values = ["x86_64"]
+  }
+
+  owners = ["099720109477"]
```
vedhavyas, line 19: Can't we move this as a variable instead ?

templates/terraform/aws/vpc/main.tf
```diff
@@ -0,0 +1,593 @@
...
+    vpc_id = try(aws_vpc_ipv4_cidr_block_association.this[0].vpc_id, aws_vpc.this[0].id, "")
+
+    create_vpc = var.create_vpc
+  }
+
+  ######################################
+  # VPC
+  ######################################
+
+  resource "aws_vpc" "this" {
```
vedhavyas, line 21: I dont think `this` is generally preferred. It would be ideal to name this instead

vedhavyas (commented):
I see only templates in this PR
Where are we planning to use this module ?

templates/terraform/aws/ec2/ami.tf
DaMandal0rian: This is the UID for canonical which is constant.

templates/terraform/aws/vpc/main.tf
DaMandal0rian: Please see https://www.terraform-best-practices.com/naming

"this" or "main" is usually used in naming convention if there is one resource created in the module.

DaMandal0rian:
> I see only templates in this PR
> Where are we planning to use this module ?

This is just a PR for the module, it will be used by all resources that we provision on AWS, which is being refactored and will be in a future PR. If I include it here, there will be many files :)

vedhavyas (approved):

PR #268, opened 2024-03-05
create rewards squid terraform declaration

DaMandal0rian:
The PR adds another squid type `rewards squid` to the explorer terraform declarations. This squid is used to process rewards separately and decouple the processing from the main squid to improve on indexing performance. The PR adds a blue/green environment for this squid.


DaMandal0rian:
This is no longer needed since the rewards squid can run on the same machine as the primary squid processor

PR #274, opened 2024-03-07
Fix IPv6 addresses associated with instances

DaMandal0rian:
Fix IPv6 address count associated with NICs in `network-primitives` module for EC2 instances. 
This fixes a regression where 2 IPv6 addresses were assigned to a NIC interface on a single AWS instance which is not needed. With this change only a single IPv6 address will be added to the instance.


vedhavyas (commented):
I'm not sure I understand this change. Can you please add more details to the PR description pls

vedhavyas (approved):

PR #276, opened 2024-03-07
CF records changes

DaMandal0rian:
This PR simply updates DNS records for subspace zone in terraform:
- add missing telemetry A records
- remove blockscout A record

### DIFF
--- a/dns/records.tf
+++ b/dns/records.tf
@@ -118,17 +118,27 @@ resource "cloudflare_record" "block_explorer" {
   zone_id = data.cloudflare_zone.subspace_network.id
 }
 
-// subspace blockscout
-resource "cloudflare_record" "blockscout" {
-  name    = "blockscout"
-  comment = "Subspace blockscout evm explorer"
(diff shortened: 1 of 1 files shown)


dns/records.tf
```diff
@@ -118,17 +118,27 @@ resource "cloudflare_record" "block_explorer" {
...
   type    = "A"
-  value   = "3.145.138.108"
+  value   = "35.89.37.220"
   zone_id = data.cloudflare_zone.subspace_network.id
 }
 
+// subspace telemetry (substation)
+resource "cloudflare_record" "substation" {
+  comment = "Subspace Telemetry (Substation)"
+  name    = "substation"
```
nazar-pc, line 135: Why do we need a substation record though?

nazar-pc (approved):

dns/records.tf
DaMandal0rian: This was requested by Justin before

PR #282, opened 2024-03-14
Fixes nginx conf to be reusable by squid modules

DaMandal0rian:
Each squid module will create it's own `server_name` directive now. This makes the config reusable by multiple squid modules for different networks.

closes #142


dnoishi (approved):
Looks good to me!

PR #415, opened 2025-02-15
removing CF modules as it will be replaced

DaMandal0rian:
These modules are unused and will be replaced with pulumi IaC in future to better manage Cloudflare LBs and omit the friction of using the terraform module for CF.


DaMandal0rian:
closes #418

PR #445, opened 2025-07-23
import cloudflare load-balancers to terraform

DaMandal0rian:
Importing the cloudflare load-balancers for our networks into terraform. It includes:

- Load-balancers
- Pools
- Monitors and Health Checks

### DIFF
--- a/resources/terraform/cloudflare-lbs/main.tf
+++ b/resources/terraform/cloudflare-lbs/main.tf
@@ -1,0 +1,287 @@
+# Cloudflare Provider Configuration
+terraform {
+  required_providers {
+    cloudflare = {
+      source  = "cloudflare/cloudflare"
+      version = "~> 4.0"
+    }
+  }
+}
+
+provider "cloudflare" {
+  api_token = var.cloudflare_api_token
+}
+
+
+# =============================================================================
+# MONITORS (Import these first)
+# =============================================================================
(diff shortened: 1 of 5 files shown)


resources/terraform/cloudflare-lbs/main.tf
```diff
@@ -0,0 +1,287 @@
...
+    prefer_ecs = "proximity"
+    mode       = "pop"
+  }
+
+}
+
+# Load Balancer 2: rpc-lb.mainnet.autonomys.xyz
+resource "cloudflare_load_balancer" "rpc_lb_mainnet" {
+  zone_id          = var.cloudflare_zone_id
+  name             = "rpc-lb.mainnet.autonomys.xyz"
```
vedhavyas, line 217: does this mean we can use `rpc-lb.mainnet.autonomys.xyz` for polkadot js or as rpc url ?

Does individual node instaces running Traefik need to generate cert for this or Cloudflare changes origin to `rpc-1.mainnet.subspace.network` ?

DaMandal0rian: That is the LB endpoint, but we use a CNAME DNS record `rpc.mainnet.autonomys.xyz` to point to it

vedhavyas (approved):

PR #454, opened 2025-07-28
update mainnet RPCs LB IPs

DaMandal0rian:
migrated RPCs and assigned previous ones as internal operators.


teor2345 (commented):
Seems fine to me, but I think @vedhavyas will have more context here.

vedhavyas (approved):

PR #466, opened 2025-09-04
Explorer auto evm chronos DNS

vedhavyas:


0xjeremyfrank (approved):
The previous ones were proxied and didn't show the ip in the `content` field which is where I had some confusion.

PR #474, opened 2025-09-16
Update proxied astral dns records post redirect setup on cloudflare

0xjeremyfrank:
This pull request updates DNS record configurations for astral related subdomains in `resources/terraform/dns/autonomys_xyz.tf`, primarily changing CNAME targets and proxy settings. These subdomains are only used to redirect to subscan which is why content was updated to `example.com`. The PR also removes an obsolete DNS record for `dev-explorer.autonomys.xyz`.

Note: The updating of the DNS records were manually done in Cloudflare in order to quickly test the redirect rule that was added. The removal of the no longer needed record was not handled manually.


vedhavyas (approved):

vedhavyas:
@jfrank-summit the changes are applied through terraform

Issue #530, opened 2026-03-11
Remove old auto-drive Terraform project after production migration


jim-counter:
## Context

PR #520 migrates Auto Drive infrastructure from the monolithic `resources/terraform/auto-drive/` project to the new module structure (`modules/auto-drive/` + `resources/terraform/auto-drive-production/`). After #520 is applied and verified, the old project directory should be removed.

## Tasks

- [x] Confirm PR #520 has been applied successfully and state is clean
- [x] Remove `resources/terraform/auto-drive/` directory
- [x] Verify no other projects reference the old directory
- [x] Update any documentation that references the old path

## Related

- PR #520

Issue #528, opened 2026-03-11
Move Auto Drive secrets out of module and into Infisical


jim-counter:
## Context

Per [vedhavyas's review on PR #520](https://github.com/autonomys/infra/pull/520#discussion_r2911779999), secrets like the RabbitMQ password should be managed via Infisical through `common.auto.tfvars` rather than generated inside the module. This aligns with the pattern used across other projects in this repo.

### 1. RabbitMQ password — `modules/auto-drive/broker.tf` lines 1-5

Currently generated inside the module via `random_password.rabbitmq_password`. This should be:
- Generated once externally (or via a one-time bootstrap)
- Stored in Infisical
- Passed into the module as a variable (e.g., `var.rabbitmq.password`)
- Added to `common.auto.tfvars` (managed by Infisical)

```hcl
# Current (in-module generation)
resource "random_password" "rabbitmq_password" {
  length           = 15
  special          = true
  override_special = "!@#$%^&*()-_=+[]{}<>:?"
}
```

### 2. RabbitMQ username — `resources/terraform/auto-drive-production/variables.tf` lines 13-18

Currently has a hardcoded default value (`"guru"`) and is marked `sensitive = true`. Should be:
- Removed from `variables.tf` default
- Stored in Infisical
- Passed via `common.auto.tfvars`

```hcl
# Current (hardcoded default)
variable "rabbitmq_username" {
  description = "RabbitMQ username"
  type        = string
  sensitive   = true
  default     = "guru"
}
```

## Tasks

- [ ] Add `rabbitmq_password` as a new field in the module's `rabbitmq` variable object
- [ ] Remove `random_password.rabbitmq_password` resource from `broker.tf`
- [ ] Update `broker.tf` line 39 to use `var.rabbitmq.password` instead of `random_password.rabbitmq_password.result`
- [ ] Remove hardcoded default from `rabbitmq_username` variable
- [ ] Add both `rabbitmq_username` and `rabbitmq_password` to `common.auto.tfvars` via Infisical
- [ ] Update `common.auto.tfvars.example` with the expected keys
- [ ] Coordinate the password value — retrieve the current password from AWS MQ console or state before switching, to avoid resetting the live broker password

## ⚠️ Migration note

The current `random_password` resource already has a value in state, so the existing broker password is stable (it doesn't regenerate on every apply). When migrating to Infisical, the current password value must be extracted from Terraform state (`terraform output` or `terraform state show`) and stored in Infisical to avoid a password rotation on the live broker.

[...]

Issue #123, opened 2023-05-02
Declarative deployment


nazar-pc:
Currently we deploy things imperatively: running scripts that modify state such that it takes desired shape. This works, but it is problematic because we can't revert changes, can't ensure state is exactly what we expect it to be and use tools designed for infrastructure creation to do so.

What we want instead is declarative deployments:
* Likely some kind of template, in case of k8s it would be Helm chart with some variables
* Then instantiation of the template with variables filled in, environment variables and secrets configured in it
* When developer wants to instantiate something, they just send a pull request, when pull request is merged, something like Argo CD will apply necessary changes to make sure deployment matches what was defined in the repository exactly
* Destruction then is as simple as reverting a pull request

For this to work we need to pick a tool for declarative deployment, key storage solution and have some templates (ideally with readme) that the rest of the team can work with somewhat autonomously.
```[tasklist]
### Tasks
- [x] Add ArgoCD manager to kubernetes
- [x] setup argo cd in cluster
- [x] create helm charts for astral and nova
- [x] create kustomize manifests for networks
- [x] update helm charts for squids rewards, account, general
- [x] create argocd path for declarative deployments of resources
```

DaMandal0rian:
#143 adds argocd manager to kubernetes cluster as an addon with terraform

DaMandal0rian:
argocd now installed in cluster

![Image](https://github.com/subspace/infra/assets/3614052/a028a0ea-3ae1-4e63-ac35-ce4b82c1b94e)

DaMandal0rian:
Added devnet k8s deployments:
- [x] https://github.com/subspace/infra/pull/307 
- [x] https://github.com/subspace/infra/pull/308
- [x] https://github.com/subspace/infra/pull/309

DaMandal0rian:
Added squid deployments for k8s:

- [x] https://github.com/subspace/infra/pull/304
- [x] https://github.com/subspace/infra/pull/305
- [x] https://github.com/subspace/infra/pull/306

DaMandal0rian:
Next agenda:
- Now i need to test these deployments in the cluster and connect it to argocd to manage the deployments when change is made to the path it tracks.
- Need to add secrets configs since sensitive info can't be used in configmaps, will look at how to have this injected from Vault. Likely solution i will use is https://github.com/kubernetes-sigs/secrets-store-csi-driver which gets the secrets from vault and mounts it as a CSI volume.
- Then i will add the secrets config to the statefulset manifests

[...]

PR #543, opened 2026-05-06
Add Terraform-managed auth Lambda (Node.js 24) with API Gateway

jim-counter:
## Summary

- Adds a new `auto-drive-auth-v2` Lambda function (Node.js 24, 512MB) and HTTP API Gateway, both in us-east-2, co-located with the DSQL cluster
- Fixes the cross-region latency issue where the previous API Gateway was in us-east-1 while the Lambda was in us-east-2
- Infrastructure has been applied — resources are live and awaiting code deployment
- Auth secrets stored in Infisical under `/auto-drive-production`
- Existing `auto-drive-auth` Lambda is untouched and continues to serve traffic until cutover

## Files changed

| File | Change |
|---|---|
| `auto-drive-production/auth.tf` | New — Lambda, IAM, API Gateway, CloudWatch resources |
| `auto-drive-production/variables.tf` | Auth-specific input variables |
| `auto-drive-production/outputs.tf` | Auth API Gateway URL and Lambda outputs |
| `auto-drive-production/providers.tf` | `aws.auth` provider alias for us-east-2 |
| `auto-drive-production/versions.tf` | Added `hashicorp/archive` provider |
| `manage.sh` | Added `auth.auto.tfvars` to auto-drive-production secrets |
| `.gitignore` | Added `*.zip` |

## Follow-up work

- [x] Deploy auth code to new Lambda and smoke test (`autonomys/auto-drive` PR to follow)
- [ ] Decommission old `auto-drive-auth` Lambda and us-east-1 API Gateway after cutover
- [ ] Evaluate auth layer multi-region strategy (DSQL replication, Lambda alias routing)
- [ ] Improve Infisical integration to support per-secret management (relates to #528)

## Related issues

- #123 (Declarative deployment)
- #530 (Remove old auto-drive Terraform project)
- #528 (Move secrets to Infisical)

## Test plan

- [x] `terraform plan` shows 10 resources to add, 0 to change, 0 to destroy
- [x] `terraform apply` completed successfully
- [x] Deploy auth code and verify health endpoint
- [x] Run auth flow smoke tests (OAuth, JWT, API key)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.tfstate*
 *.tfplan
 *.tfvars
+*.zip
 config.toml
 proxied.json
 .DS_Store
@@ -21,3 +22,4 @@ resources/mainnet/keystore/*
 /.idea
 /modules/node-utils/target
 /modules/node-utils/data
+.env
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 *.tfstate*
 *.tfplan
 *.tfvars
-*.zip
 config.toml
 proxied.json
 .DS_Store
--- a/resources/terraform/auto-drive-production/auth.auto.tfvars.example
+++ b/resources/terraform/auto-drive-production/auth.auto.tfvars.example
@@ -1,0 +1,4 @@
+auth_jwt_secret            = ""
+auth_api_secret            = ""
+auth_dsql_cluster_endpoint = ""
+auth_dsql_cluster_arn      = ""
--- a/resources/terraform/auto-drive-production/auth.tf
+++ b/resources/terraform/auto-drive-production/auth.tf
@@ -25,10 +25,10 @@ locals {
 
   # Non-sensitive operational config, hardcoded to keep Infisical lean.
   # Only actual secrets and infrastructure identifiers go in auth.auto.tfvars.
-  auth_cors_allowed_origins                    = "*"
-  auth_jwt_secret_algorithm                    = "RS256"
-  auth_log_level                               = "info"
-  auth_revoke_token_emitted_before_in_seconds  = 1746187810
+  auth_cors_allowed_origins                   = "*"
+  auth_jwt_secret_algorithm                   = "RS256"
+  auth_log_level                              = "info"
+  auth_revoke_token_emitted_before_in_seconds = 1746187810
 }
 
 # Minimal placeholder used only on the first `terraform apply` to satisfy the
--- a/resources/terraform/auto-drive-production/auth.tf
+++ b/resources/terraform/auto-drive-production/auth.tf
@@ -1,0 +1,179 @@
+################################################################################
+# Auth Lambda + HTTP API Gateway
+#
+# All resources use provider alias `aws.auth` (us-east-2), co-located with
+# the Aurora DSQL cluster. This replaces the manually-deployed auto-drive-auth
+# Lambda (nodejs20.x, us-east-2) and its API Gateway (previously misconfigured
+# in us-east-1, causing cross-region latency on every auth request).
+#
+# Code deployments are handled separately from infrastructure changes:
+#   yarn workspace auth lambda:build
+#   zip -j apps/auth/build/index.zip apps/auth/build/index.js
+#   aws lambda update-function-code \
+#     --function-name auto-drive-auth-v2 \
+#     --zip-file fileb://apps/auth/build/index.zip \
+#     --region us-east-2
+################################################################################
+
+locals {
+  auth_name = var.auth_function_name
+  auth_tags = {
+    Name    = local.auth_name
+    Project = "auto-drive"
+    Component = "auth"
+  }
+}
+
+# Minimal placeholder used only on the first `terraform apply` to satisfy the
+# requirement that a Lambda function must have code at creation time.
+# Subsequent code updates are done via `aws lambda update-function-code`.
+data "archive_file" "auth_placeholder" {
+  type        = "zip"
+  output_path = "${path.module}/.auth-placeholder.zip"
+  source {
+    content  = "exports.handler = async () => ({ statusCode: 503, body: JSON.stringify({ message: 'deploying' }) })"
+    filename = "index.js"
+  }
+}
+
+################################################################################
+# IAM
+################################################################################
+
+resource "aws_iam_role" "auth_lambda" {
+  provider = aws.auth
+  name     = "${local.auth_name}-role"
+
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Effect    = "Allow"
+      Principal = { Service = "lambda.amazonaws.com" }
+      Action    = "sts:AssumeRole"
+    }]
+  })
+
+  tags = local.auth_tags
+}
+
+resource "aws_iam_role_policy_attachment" "auth_basic_execution" {
+  provider   = aws.auth
+  role       = aws_iam_role.auth_lambda.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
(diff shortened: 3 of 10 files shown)


resources/terraform/auto-drive-production/auth.tf
```diff
@@ -0,0 +1,186 @@
...
+resource "aws_iam_role_policy" "auth_dsql_access" {
+  provider = aws.auth
+  name     = "${local.auth_name}-dsql-access"
+  role     = aws_iam_role.auth_lambda.id
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Effect   = "Allow"
+      Action   = ["dsql:DbConnect", "dsql:DbConnectAdmin"]
```
vedhavyas, line 82: `dsql:DbConnectAdmin` is admin-level DSQL access (lets the connection bypass row-level auth grants, run schema ops, etc.). Auth services typically only need `dsql:DbConnect` to authenticate the Lambda's IAM identity when opening a connection. Worth checking whether the auth code actually exercises admin paths against the cluster. If not, tightening this to just `dsql:DbConnect` is a free win.

resources/terraform/auto-drive-production/auth.tf
```diff
@@ -0,0 +1,186 @@
...
+    Project   = "auto-drive"
+    Component = "auth"
+  }
+
+  # Non-sensitive operational config, hardcoded to keep Infisical lean.
+  # Only actual secrets and infrastructure identifiers go in auth.auto.tfvars.
+  auth_cors_allowed_origins                    = "*"
+  auth_jwt_secret_algorithm                    = "RS256"
+  auth_log_level                               = "info"
+  auth_revoke_token_emitted_before_in_seconds  = 1746187810
```
vedhavyas: This is operational state (the "revoke all tokens issued before" epoch) baked into Terraform locals. Bumping it requires a `terraform apply` rather than a config flip, and the audit trail lives only in `git log` on this file. Was the thinking to make incident response one-step (single apply does it), or would moving it to SSM Parameter Store / Secrets Manager and reading at Lambda cold-start be cleaner long-term? The 2025-05-02 value also looks stale, wondering if it has been touched since the original write.

resources/terraform/auto-drive-production/auth.tf
```diff
@@ -0,0 +1,186 @@
...
+  handler          = "index.handler"
+  runtime          = "nodejs24.x"
+  memory_size      = var.auth_memory_size
+  timeout          = var.auth_timeout
+  filename         = data.archive_file.auth_placeholder.output_path
+  source_code_hash = data.archive_file.auth_placeholder.output_base64sha256
+
+  environment {
+    variables = {
+      JWT_SECRET                             = var.auth_jwt_secret
```
vedhavyas, line 114: `JWT_SECRET` and `API_SECRET` land in plaintext on the Lambda configuration, visible to anyone with `lambda:GetFunctionConfiguration` and reproduced in CloudTrail. I can see the project pattern is Infisical to tfvars to env vars and this matches that, so not a blocker. Worth a thought down the line if the IAM surface widens: KMS-encrypted env vars with a customer-managed key, or fetch from Secrets Manager at cold-start, are the usual next steps.

resources/terraform/auto-drive-production/auth.tf
```diff
@@ -0,0 +1,186 @@
...
+# Explicit log group so we control retention.
+# Without this, Lambda auto-creates it with no retention policy.
+resource "aws_cloudwatch_log_group" "auth_lambda" {
+  provider          = aws.auth
+  name              = "/aws/lambda/${local.auth_name}"
+  retention_in_days = var.auth_log_retention_days
+  tags              = local.auth_tags
+}
+
+resource "aws_lambda_function" "auth" {
```
vedhavyas, line 101: Just confirming, does the auth Lambda reach DSQL over its public endpoint (default), or does the DSQL cluster sit behind a private VPC endpoint? If the latter, the Lambda needs `vpc_config` attached to the right subnets plus an SG allowed by DSQL, otherwise the `dsql-signer` calls will hang on first invoke. Nothing in the diff configures one, so wanted to make sure that is intentional.

resources/terraform/auto-drive-production/auth.auto.tfvars.example
```diff
@@ -0,0 +1,4 @@
+auth_jwt_secret            = ""
+auth_api_secret            = ""
```
vedhavyas, line 2: be consistent with other projects. 
Are these common for all user if so, then it should be common.auto.tfvars.example. 
else user.**

.gitignore
```diff
@@ -2,6 +2,7 @@
 *.tfstate*
 *.tfplan
 *.tfvars
+*.zip
```
vedhavyas: what does zip contents we are using for ?
Might be ideal to add that to your git/info/exclude if this is your specific gitignore

vedhavyas (commented):
Read through end-to-end. The shape is clean. Lambda plus HTTP API Gateway v2 in us-east-2, IAM scoped to the DSQL cluster ARN, explicit CloudWatch log group with retention, and lifecycle.ignore_changes on the code fields so the deploy pipeline doesn't fight Terraform. A few observations inline. The DSQL admin scope and the revoke-epoch-in-locals are the two I'd flag as worth thinking about; the rest are softer.

resources/terraform/auto-drive-production/auth.tf
jim-counter: Great point. Needs a code change but definitely worth doing. Issue raised: https://github.com/autonomys/auto-drive/issues/699

resources/terraform/auto-drive-production/auth.tf
jim-counter: Yes, I suspect this is a value that was introduced to handle a single, particular scenario.

From an architectural perspective your suggestion to use Parameter Store feels like the right call. We can use that insight to inform future development and any refactor of the auth lambda code.

resources/terraform/auto-drive-production/auth.tf
jim-counter: Agreed - that's a better approach which we can adopt when the time is right.

resources/terraform/auto-drive-production/auth.tf
jim-counter: Intentional - the Lambda uses DSQL's public endpoint and IAM tokens, no VPC config needed at this stage.

resources/terraform/auto-drive-production/auth.auto.tfvars.example
jim-counter: Ha! I agonised over this one when I made my standardisation pass.

My reasoning is that the key difference is that auto-drive-production is a multi-component project (auth, backend, frontend, gateway, etc.) unlike every other project which is single-component. Using `common.auto.tfvars` would work now with just auth, but as soon as backend is added we'd have one massive file mixing unrelated secrets. Component-specific naming (`auth.auto.tfvars`, future `backend.auto.tfvars`) scales better.

If you have an alternative convention for a multi-component setup, happy to hear it.

.gitignore
jim-counter: Yep, this was sloppy. Addressed in https://github.com/autonomys/infra/pull/543/commits/f7ec4a64c7bcd6da45fb3cb64241e41036760009.

vedhavyas (approved):
Re-reviewed at 386e02a. All my earlier points are addressed or tracked as follow-ups (DSQL admin scope in auto-drive#699, revoke-epoch deferred to a Parameter Store refactor). validate passes and I pushed a small fmt alignment on the auth locals. Good to land.

PR #564, opened 2026-06-17
Manage the uptime-kuma deploy from terraform

vedhavyas:
Brings the telemetry box's uptime-kuma under terraform, the same way the other services work: its docker-compose.yml now lives in the repo and a null_resource copies it to the box and runs `docker compose up -d`, triggered on `filesha256(compose)` so editing the compose and applying redeploys it.

The connection uses the SSH agent rather than a key file, since the box's deploy key is passphrase-protected (terraform can't read an encrypted key directly). The first apply was a no-op, Kuma stayed up the whole time (verified, still at its original uptime).

substrate-telemetry is intentionally left as-is for now: it builds its images from a cloned source fork on the box rather than pulling a registry image, so managing it cleanly is a separate piece of work.

### DIFF
--- a/resources/terraform/telemetry/deploy.tf
+++ b/resources/terraform/telemetry/deploy.tf
@@ -15,8 +15,11 @@ resource "null_resource" "deploy_uptime_kuma" {
     timeout = "120s"
   }
 
+  # on a fresh/replaced box: wait for cloud-init and install docker if missing (no-op on the existing box)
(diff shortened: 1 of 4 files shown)


vedhavyas:
bugbot run

vedhavyas:
bugbot run

vedhavyas:
bugbot run

PR #581, opened 2026-08-04
replace new relic with node-exporter, fix vmagent remote-write [blocked on vmetrics rebuild]

vedhavyas:
PARKED until the vmetrics box rebuild lands. Not ready to merge.

Replaces New Relic with node-exporter and fixes the vmagent remote-write, which has been failing fleet-wide because it pushes to port 8428 and Cloudflare never forwards that.

The reason this waits: we're rebuilding the metrics box from scratch (it's unreachable, nobody holds a key) and putting vmauth in front as a single authenticated gateway for both metrics and logs. That very likely changes the endpoint and the auth model this PR hardcodes, and nodes will want vector alongside vmagent to ship logs. Merging now would mean rolling every node twice.

Once the new box is up, I'll update the vmagent URL and auth here, add vector, and then roll it out in one pass.

New Relic removal and node-exporter stand on their own and shouldn't need rework.


vedhavyas:
Closing - we're not rebuilding the vmetrics box, so the vmagent remote-write fix and node-exporter this depended on aren't going anywhere. The New Relic removal is still worth doing on its own and I'll raise that separately if we want it.

